PR #25, opened 2015-02-04
OS X Target

remaerd:
Hello, I added the OS X Target into CryptoSwift.


krzyzanowskim:
I'll add OSX target here #136

PR #33, opened 2015-02-27
Eliminating Swift 1.2 warnings

winzig:
Apparently no “return nil” is needed after an assertionFailure.


krzyzanowskim:
This change will break Release build where assertionFailure is noop

winzig:
Ahhh, whoops! :)

Issue #16, opened 2015-01-20
only support iOS 8+


spritesun:
Request to support iOS 7 as well.

krzyzanowskim:
Now I think it was bad decision. Will it ever work with 7.1 when embedded dylibs/frameworks only run on iOS 8 or later?

krzyzanowskim:
According to discussion here https://github.com/CocoaPods/CocoaPods/issues/3193 I can't have deployment target set to version 7.1 any longer.

dkhamsing:
Man this is a bummer

krzyzanowskim:
You have to handle it by yourself by copying files to your project, not using framework, because framework is the limitation here.

dkhamsing:
So your advice is, copy the files if supporting ios7?

krzyzanowskim:
you can copy, or just add to your project, and yes this seems the only way to do so.

dkhamsing:
thx

PR #35, opened 2015-03-06
Added iPhone OS deployment target for iOS 7.0.

karupanerura:
All tests is successful on iOS7 simulator.


krzyzanowskim:
yes, but framework itself can't be used on iOS < 8.0 Check this: https://github.com/krzyzanowskim/CryptoSwift/issues/16

krzyzanowskim:
won't be merged

PR #46, opened 2015-03-30
[FIX] Memory leak with unsafe calloc

yannickl:
Hi,

I found a memory leak on your library while debugging my app. If I targeted the issue correctly (and well understood your lib), the leak concerns an unsafe pointer over a `calloc` which is not deallocated in the `prepare` method of the `HashBase` class.

Thank you for your great lib. ;)

Best,

Yannick


krzyzanowskim:
nice find. 👍

PR #59, opened 2015-05-06
Mac port

MaddTheSane:
Also various changes.


krzyzanowskim:
I would prefer if you split these commits in to (at least) pull requests, as they touch different part of codebase, not just OSX target.

krzyzanowskim:
and how that relate to #25 ?

MaddTheSane:
I wasn't aware of pull request #25 when I made my pull request.

Also, I will split up my changes into different pull requests when I get the chance.

PR #67, opened 2015-05-15
Adds an aggregate target which runs a script to build a universal fra…

jagreenwood:
…mework.  The script currently copies the framework to the `Framework` dir


krzyzanowskim:
thanks!

PR #69, opened 2015-05-16
ios 7 compatibility changes

pieisgood:
Added a folder containing ios 7 compatible changes, these can be reduced to only the files that changed.


krzyzanowskim:
Thank you, this look interesting. However, I will pass on that. There will be to much efforts to maintain two versions of the same codebase. 

Is there another way to make it ios7 compatible, without making a copy ?

pieisgood:
I can look into that, it's likely possible but it would necessarily remove the CryptoSwift namespace for classes that inherit from other CryptoSwift classes. Not sure you want to do that either.

krzyzanowskim:
pass with that one. Thanks for you work anyway, appreciate!

Issue #71, opened 2015-05-30
String extension `.md5()` -- should it return lowercase?


LoganWright:
If I run crypto on identical strings like so (a is CommonCrypto, b is CryptoSwift): 

```Swift
        let string = "somerandomstringtohash"
        let a = hashable.md5
        let b = hashable.md5()!
        println("a: \(a)")
        println("b: \(b)")
```

I get a readout like so: 

>a: 7c9c695f659046b137f36aa9507e7730
b: 7C9C695F659046B137F36AA9507E7730

-----

Now, I am not the most up to date on security measures, but I need an md5 has for the marvel api, and it only works with lowercase strings.  I get around this by making it lowercase, but I wanted to point it out for you in case it's something that should be fixed.  If it's supposed to be this way, just ignore me :)

Thanks!

krzyzanowskim:
actually it doesn't matter. It's just a string representation of bytes. Sometimes people use lowercase, other time uppercase, sometimes It is preceded by "0x". Although... it may be that lowercase representation is more popular, in this case yes, we should change it. Will you prepare pull request for that ?

DaveWoodCom:
Personally, I'd be cautious about implementing this change since it could break existing users of the library. Any one using the library already that is expecting it to be lower case for comparisons will have already converted the case in their code.

krzyzanowskim:
yea me too.
On the other hand better change it now than later. I did quick check, and seems that lowercase string is more popular.

ctews:
+1 for changing this, had the same troubles here! :)

alessiomaffeis:
Stumbled into the same problem too... +1

krzyzanowskim:
There is PR https://github.com/krzyzanowskim/CryptoSwift/pull/72 but it's not ready to merge. If not updated soon I'll apply this change by myself.

krzyzanowskim:
The work is done.

alessiomaffeis:
🎉

robinvpersie:
I got the same problem, how to get the upper, thanks

PR #72, opened 2015-05-30
hex string returns lowercase

LoganWright:
I went into the original `toHexString` function and added the lowercase call there.  Again, I'm very new to security, so I'm not sure how this will effect the other hashes.  Let me know if this works!

krzyzanowskim/CryptoSwift#71

### DIFF
--- a/CryptoSwift/NSDataExtension.swift
+++ b/CryptoSwift/NSDataExtension.swift
@@ -95,9 +95,9 @@ extension NSData {
         
         var s:String = "";
         for byte in bytesArray {
-            s = s + String(format:"%02X", byte)
+            s = s + String(format:"%02x", byte)
         }
-        return s.lowercaseString;
+        return s
     }
     
     public func arrayOfBytes() -> [UInt8] {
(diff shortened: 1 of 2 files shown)


michaelmcv:
I had to use the lowercaseString workaround on the sha256() function result in order to connect to the api I am authenticating against, so it made a difference there as well. Perhaps it should be added to all the hashing functions in this library.

krzyzanowskim:
related #71 It's just String representation of data.

krzyzanowskim:
I any case, this pull request will fail on tests, please update with tests

krzyzanowskim:
@LoganWright can you add/update tests for this change?

LoganWright:
@krzyzanowskim Sorry about taking so long, tests have been updated!

krzyzanowskim:
@LoganWright would you mind prepare new PR against `swift20` branch? I'm not going to merge it to `master` but willing to apply this change to next version

krzyzanowskim:
Instead of call lowercaseString, it's enough to change format from `%02X` to `%02x`

LoganWright:
@krzyzanowskim I just updated with your suggestion and moved the pr to the appropriate branch, #112.  Feel free to close this one

PR #84, opened 2015-06-18
Remove CryptoSwift namepsace for iOS7

zhupeijun:
This will remove CryptoSwift namepsace and can be imported to iOS7 project.

1. Use type alias to different the enum function and class function.
2. User third function to different Int shiftLeft and Generic shiftLeft.
3. Remove HashBases namespace.


krzyzanowskim:
This is great work, however I decided not merge this changes to the main repo.
Thanks.

PR #92, opened 2015-07-09
Meeting Xcode Beta3 requirements

Sali0m:
With the latest beta of xcode T has been changed to Element


krzyzanowskim:
Why this is a thing at all is beyond me ;)

Issue #57, opened 2015-04-28
AES decryption with PKCS7 padding seems to read last byte incorrectly


gyamana:
Hello,

A newbie iOS developer here, but please hear me out.

I have been scratching my head over the last couple of hours around trailing characters after decryption.

I am using CryptoSwift 0.0.10, and it seems that the PKCS7.remove() function is removing one byte of padding only, despite the bytes value being larger than 1.

The following is the byte array before padding removal:

    [84, 104, 105, 115, 32, 105, 115, 32, 109, 121, 32, 111, 114, 105, 103, 105, 110, 97, 108, 32, 116, 101, 120, 116, 8, 8, 8, 8, 8, 8, 8, 8]

The following is the output I see after padding removal has been executed:

    This is my original text

As you can see, there are 7 special characters trailing the decrypted text (the characters show up as 'backspace', which corresponds to U+0008) 

I have managed to work out that I can actually change the CryptoSwift's code under the Pods directory and recompile to see the results.

Initially I put in a series of println statements to grab the above.

The code seems to be correct, and it actually works if I cut and paste the first line like so:

    var padding = Int(bytes.last!) // last byte
    padding = Int(bytes.last!) // last byte

However without re-initializing the padding variable, it remains as a value of 1 no matter what the value of the last byte is.

Would you be able to look at whether this is an issue, or a misuse of the library?

Regards,

George

krzyzanowskim:
I don't see issue here. eight 8 at the end says that remove() should remove eight bytes, so it remove them.

```swift
public func remove(bytes: [UInt8], blockSize:Int? = nil) -> [UInt8] {
    var padding = Int(bytes.last!) // last byte

    if padding >= 1 {
        return Array(bytes[0..<(bytes.count - padding)])
    }
    return bytes
}


let bytes = remove([84, 104, 105, 115, 32, 105, 115, 32, 109, 121, 32, 111, 114, 105, 103, 105, 110, 97, 108, 32, 116, 101, 120, 116, 8, 8, 8, 8, 8, 8, 8, 8], blockSize: 16)

bytes.last // = 116
```

gyamana:
Hmm, that is puzzling..

This only occurs when I grab the version of the library from CocoaPods.

I hadn't picked this issue up initially since the trailing characters weren't visible with println, but my colleague raised the issue and I am able to produce the same.

I am calling the library like so:

[...]

PR #98, opened 2015-07-17
Fix issue where grabbing Int from UInt8 directly break the int 

krummler:
There seems to be some issue where getting the last (optional) item from the bytes array, causes a garbled integer.

By wrapping the last call in a optional check, it also fixes a possible crash when the bytes array is empty.

### DIFF
--- a/CryptoSwift/PKCS7.swift
+++ b/CryptoSwift/PKCS7.swift
@@ -35,11 +35,12 @@ public struct PKCS7: Padding {
     }
     
     public func remove(bytes: [UInt8], blockSize:Int? = nil) -> [UInt8] {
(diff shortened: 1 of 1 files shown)


krzyzanowskim:
Byte array cant be empty for pkcs7. Garbled integer was a swift optimizer bug, visible in debug builds only, I thik it was addressed already

krzyzanowskim:
https://github.com/krzyzanowskim/CryptoSwift/issues/57

krzyzanowskim:
As I said before, the problem was addressed earlier in #57. More, this bug has been fixed in upcoming Swift 2.0, no need to introduce more changes here.

PR #107, opened 2015-07-30
Added Carthage to installation section

skyylex:
Carthage is another great posibillity to build CryptoSwift. So I've added instructions about it.


krzyzanowskim:
thanks

PR #121, opened 2015-09-04
Added conditional on init to AES avoid unintentionally returning nil

foobar8675:
my call to creating the AES object was returning nil when it looked like it shouldn't be. so i just added a check for it.


krzyzanowskim:
great, want to add test for that case? (please check whitespaces before commit)

foobar8675:
hi krzyzanowskim, just added a unit test for that case, taking your key and iv values and turning them into strings.

PR #125, opened 2015-09-17
Changed code signing identity to revered "iPhone Developer"

JaviSoto:
Fixes Carthage build issues.


krzyzanowskim:
Can't find now, but I think there was a PR changing it to the state it is now, and it was due to Carthage. I'm so confused with Carthage support TBH.

The question is, how does it affect OSX builds?

JaviSoto:
I haven't touched the OS X configuration.

Carthage is just going to build the target using the signing identity that's set up by default. It's *way* more likely that people will have a development certificate (which doesn't need a provisioning profile that matches the bundle ID), than to have a production certificate. This is how all the Carthage projects I use are set up :)

JaviSoto:
Would you mind getting this merged? :)

krzyzanowskim:
let's recap, what is wrong with this config:

screen shot 2015-09-25 at 11 27 40

krzyzanowskim:
Follow up https://twitter.com/krzyzanowskim/status/647392670999015424

PR #126, opened 2015-09-19
set xcode project to allow App Extension API only

kkpoon:
set the xcode project to App Extension API only, mark safety for App Extension use.


kkpoon:
@krzyzanowskim Hello, would you accept this pull request. This configuration in Xcode project could remove the warning while I use your library in a iOS App Extension, for example, Today Extension.

Thank you

PR #159, opened 2015-11-13
Rabbit stream cipher

dkalachov:


krzyzanowskim:
👍🙏

CryptoSwift/Rabbit.swift
```diff
@@ -0,0 +1,206 @@
...
+    
+    private func nextState() {
+        // Before an iteration the counters are incremented
+        var carry = p7
+        for var j = 0; j < 8; j++ {
+            let prev = c[j]
+            c[j] = prev &+ a[j] &+ carry
+            carry = prev > c[j] ? 1 : 0 // detect overflow
+        }
+        p7 = carry // save last carry bit
```
krzyzanowskim: looks like p7 could be returning value and parameter for nextState(), wdyt?

dkalachov: I think that's not necessary. 

`p7` is a part of the cipher state which consists of 513 bits (`x[0..7] x 32bit`, `c[0..7] x 32bit` and `p7`) and there is no need to pass it explicitly in `nextState()` method.

PR #172, opened 2015-12-04
Added seed ability to crc

richlowenberg:
Let me know what you think of this - Hash.swift feels a little messy to me, as it uses a tuple and can't set a default parameter. There may be a better way. Other than that everything is handled by default parameters. 

In our organization providing a CRC seed is pretty important - don't know how critical that is for other projects though.

Thanks for a great lib!


krzyzanowskim:
Thanks!
I've should remove CRC from Hash.swift at all, it's not hash after all.

PR #191, opened 2016-01-22
don't set ONLY_ACTIVE_ARCH to YES

emaloney:
When embedding this project directly into another Xcode project in a target that:

• Is being built for a platform that has a simulator/device dichotomy (iOS, tvOS, watchOS), and
• Is inside a parent project building a Debug configuration

You will get one of two kinds of errors:

• Either the framework module itself is missing at the 'import' level, or
• Some or all of the framework's public symbols can't be found

The solution is to disable ONLY_ACTIVE_ARCH, which will ensure that both processor architectures are built and the symbols can be found.


emaloney:
If you have any questions about this, please let me know & I can show you how to trigger the problem I hit that this PR fixes.

krzyzanowskim:
ONLY_ACTIVE_ARCH shout be YES only for Debug.
Compared this setup with Alamofire...

emaloney:
I have re-tested using Xcode 7.3 beta 2 and the issue that required this change previously is no longer occurring.

PR #192, opened 2016-01-22
Full bitcode is needed for running Debug on an Apple TV device

emaloney:
Without this setting, it won't be possible to run a Debug build on the device; 'marker' only works in the sim


krzyzanowskim:
thank you

PR #196, opened 2016-01-26
Faster compilation times (60% faster)

### LINKED ISSUE #195: Slow compilation times
Hi Marcin,

as you know, I'm a big fan of CryptoSwift and I'm using it in [Buildasaur](https://github.com/czechboy0/Buildasaur). However, I've noticed that compiling CryptoSwift seems orders of magnitude slower than any other Pod I use. I ran the compilation profiler described [here](http://irace.me/swift-profiling/) and here's a gist with the results (sorted from slowest): https://gist.github.com/czechboy0/316b7293ceea8388496b

You can see that `func toUInt32Array(slice: ArraySlice<UInt8>) -> Array<UInt32>` in `Utils.swift` takes almost **10 seconds** to compile, followed by decryption in AES with over 6 seconds. The next ones drop down to hundreds of milliseconds, which is still pretty high but not a dealbreaker. 

If you could fix these two, it'd decrease the compilation time by ~15 seconds, which I'd definitely appreciate (as would likely everyone else).

Let me know your thoughts or whether I should take a look at it myself.

czechboy0:
Broke up complicated expressions which were taking ages to compile.

See gist of compilation times before and after this change here: https://gist.github.com/czechboy0/316b7293ceea8388496b

This improved the compilation time from 22 seconds to 9 seconds (60% better), but can be improved even further by breaking up expressions in the (now) slowest functions, still taking hundreds of milliseconds to compile.

Mostly fixes #195.

(All tests passed locally.)

### DIFF
--- a/Sources/CryptoSwift/AES.swift
+++ b/Sources/CryptoSwift/AES.swift
@@ -139,10 +139,29 @@ final public class AES {
             t[2] = b[2] ^ rk[r][2]
             t[3] = b[3] ^ rk[r][3]
             
-            b[0] = T0[Int(t[0] & 0xFF)] ^ T1[Int((t[1] >> 8) & 0xFF)] ^ T2[Int((t[2] >> 16) & 0xFF)] ^ T3[Int(t[3] >> 24)]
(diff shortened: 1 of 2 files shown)


hamchapman:
This is awesome :+1:

krzyzanowskim:
Yep. I'm aware of that. It slow down with latest Swift though. Now we won't see the improvement in Swift 2.3 hahha

I will merge it, just have to look inside ;)

czechboy0:
:+1:

PR #200, opened 2016-01-29
fix some new deprecation warnings

emaloney:
this is some of the low-hanging fruit; 'Bit' and C-style for loops still generate warnings under the Xcode 7.3 beta

### DIFF
--- a/Sources/CryptoSwift/Rabbit.swift
+++ b/Sources/CryptoSwift/Rabbit.swift
@@ -64,12 +64,12 @@ final public class Rabbit {
         
         // Key divided into 8 subkeys
         var k = [UInt32](count: 8, repeatedValue: 0)
-        for var j = 0; j < 8; j++ {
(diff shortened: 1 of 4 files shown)


Sources/CryptoSwift/Rabbit.swift
```diff
@@ -64,12 +64,12 @@ final public class Rabbit {
         
         // Key divided into 8 subkeys
         var k = [UInt32](count: 8, repeatedValue: 0)
-        for var j = 0; j < 8; j++ {
+        for var j = 0; j < 8; j += 1 {
```
krzyzanowskim: for (;;) is deprecated as well, isn't it ?

emaloney: It is. However, I left those and the `Bit`-related deprecations alone, since those may have different solutions depending on your design aesthetic.

PR #215, opened 2016-02-17
Refactored SHA2 final hash generation to solve "Expression was too complex" error.

oscart:
- [FIXED] Error: "Expression was too complex to be solved in reasonable time; consider breaking up the expression into distinct sub-expressions"

### DIFF
--- a/Sources/CryptoSwift/SHA2.swift
+++ b/Sources/CryptoSwift/SHA2.swift
@@ -273,8 +273,13 @@ final class SHA2 : HashProtocol {
         result.reserveCapacity(hh.count / 4)
         variant.resultingArray(hh).forEach {
             let item = $0.bigEndian
-            result += [UInt8(item & 0xff), UInt8((item >> 8) & 0xff), UInt8((item >> 16) & 0xff), UInt8((item >> 24) & 0xff),
-                       UInt8((item >> 32) & 0xff),UInt8((item >> 40) & 0xff), UInt8((item >> 48) & 0xff), UInt8((item >> 56) & 0xff)]
+            var chunk = [UInt8]()
+            chunk.reserveCapacity(8)
+            for i in 0..<8 {
+                let shift = UInt64(8 * i)
+                chunk.append(UInt8((item >> shift) & 0xff))
+            }
+            result += chunk
         }
         return result
     }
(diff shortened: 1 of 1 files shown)


Sources/CryptoSwift/SHA2.swift
```diff
@@ -273,8 +273,13 @@ final class SHA2 : HashProtocol {
         result.reserveCapacity(hh.count / 4)
         variant.resultingArray(hh).forEach {
             let item = $0.bigEndian
-            result += [UInt8(item & 0xff), UInt8((item >> 8) & 0xff), UInt8((item >> 16) & 0xff), UInt8((item >> 24) & 0xff),
-                       UInt8((item >> 32) & 0xff),UInt8((item >> 40) & 0xff), UInt8((item >> 48) & 0xff), UInt8((item >> 56) & 0xff)]
+            var chunk = [UInt8]()
+            chunk.reserveCapacity(8)
```
krzyzanowskim: this is fine, really, though, how about `chunk[i] = ... ` ? and rename `chunk` to `partialResult` or something more descriptive.

krzyzanowskim:
where did you get the "error" from compiler about too complex expression? what version of Swift ?

oscart:
I'm currently using: Apple Swift version 3.0-dev (LLVM a7663bb722, Clang 4ca3c7fa28, Swift 1c2f40e246)

I installed it yesterday when I was trying to install [kylef/JSONWebToken.swift](https://github.com/kylef/JSONWebToken.swift) (which uses CryptoSwift as a dependency) with the Swift Package Manager.

I got this error and also a couple of warnings.

krzyzanowskim:
Thank you. This seems to be Swift regression #216

Sources/CryptoSwift/SHA2.swift
oscart: I couldn't use the square brackets notation `chunk[i] = ...` there. I got an "Index out of range" error. It seems we can't create an Array that has pre-allocated memory but does not contain elements: http://stackoverflow.com/questions/24427905/array-with-capacity-in-swift/24428086#24428086

with the `reserveCapacity` call `append` should behave better, though.

krzyzanowskim: no, but you can create array with default values ;) nvm.

[UInt8](count: 8, repeatedValue: 0)  

--  
Marcin

oscart: yes, I though that as well, but I wasn't sure if creating the array with defaults would performs better than just reserving the capacity and appending. Thank you.

PR #250, opened 2016-04-26
Fixed Bug

McClementine:
Added 'import Foundation' to prevent build fail when running 'swift build'


krzyzanowskim:
Foundation is not used in this file, at all.
It must be something else.

McClementine:
@krzyzanowskim NSData fails with "unresolved identifier 'NSData' ", which is fixed by importing foundation when using the 'open-source' version of swift. (I'm on Linux)

krzyzanowskim:
but.... there is not NSData in this file

McClementine:
@krzyzanowskim This is what my output is telling me

McClementine:
@krzyzanowskim wrong file, will close and add new

PR #254, opened 2016-04-29
Add a public initializer to `NoPadding`

cyhsutw:
**What's this PR for?**

Add a public constructor to struct `NoPadding` to fix compilation error when using `NoPadding()` outside the module.

**What's going on?**

While using the example code in `README`:

> [AES without data padding](https://github.com/krzyzanowskim/CryptoSwift#aes)

```swift
let encrypted: [UInt8] = try! AES(key: "secret0key000000", iv:"0123456789012345", blockMode: .CBC, padding: NoPadding()).encrypt(input)
```

The code won't compile and the complier is emitting this error:

```
'NoPadding' cannot be constructed because it has no accessible initializers
```

### DIFF
--- a/Sources/CryptoSwift/NoPadding.swift
+++ b/Sources/CryptoSwift/NoPadding.swift
@@ -7,6 +7,10 @@
 //
 
 public struct NoPadding: Padding {
+    public init() {
+    
+    }
+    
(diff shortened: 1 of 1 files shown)


krzyzanowskim:
👍 good job!

PR #259, opened 2016-05-13
Added reflection control to CRC-32

lm2s:
Added reflection control to CRC-32 (calculation can now reflect or not the bytes and result).

Removed UInt32 reversing util function (reverseBytes(_:)) which produced wrong results and added a reverse function for UInt8 and UInt32.


krzyzanowskim:
thank you @lm2s great work!

Issue #284, opened 2016-07-23
Allow to set initial counter explicitly.


skeeet:
This would permit direct access to any block without having to compute the previous ones. 
I'll gladly add this feature, but have some API considerations:
Option 1: Add new UpdatableStreamingCryptor protocol, similar to UpdatableCryptor, but with specific block counter in all methods. 
Option 2: Add new StreamCipher protocol (derived from Cipher), containing counter-aware methods
Option 3: Add new stuff exclusively to ChaCha20 implementation.

Personally falling into first option, but it is hard to generalize interface for all other random-access ciphers.

krzyzanowskim:
Check how AES supports CTR with UpdatableCryptor (There is CTRModeWorker used internally) - counter is not publicly available. You need it randomly accessible to archive what?

skeeet:
I heed it to be randomly accessible to use on encrypted media (videos/images). Say we have 10gb encrypted video file. Using current UpdatableCryptor interface, In order to fast forward to the middle, we'll have to read whole 5gb and just to rewind internal counter to proper state, and repeat whole process if user decided to rewind few seconds back.

krzyzanowskim:
It could be parameter for `makeEncryptor()` from `Cryptors`. Since it's not general for every possible block mode it may be defined outside the protocol, per cipher.

xlc:
I have a very similar requirements but for CTR mode. I need a way to random access decrypted data from encrypted videos file to allow user to seek.

Basically I want to somehow modify the counter variable in `CTRModeWorker` directly from the update method.

krzyzanowskim:
I renamed title of this issue to more general.
This functionality is now in `RandomAccessCryptor` protocol and is implemented for AES (for now).

krzyzanowskim:
This feature is implemented (`swift3` branch) and will be part of next major release.
Feel free to reopen if you find it not working.

PR #289, opened 2016-07-28
fix AES-CTR mode with incremental update

xlc:
Fixes #287

PR #290, opened 2016-07-28
Allow seek to any position with CTR mode

xlc:
I need this in order random access decrypted data from encrypted videos file to allow user to seek. This can be a starting point of #284 

I don't feel this API is very clean. Open to suggestions.


krzyzanowskim:
Thanks for great tests and inspiration. I reimplemented it in slightly different way along with fix to AES-CTR https://github.com/krzyzanowskim/CryptoSwift/pull/289#issuecomment-236043379

PR #282, opened 2016-07-16
Disable Bitcode for Xcode 8

colemancda:

Issue #240, opened 2016-04-16
Swift 3.0 support - bitcode support


colemancda:
I created a `Swift3` branch on my fork and am working on Swift 3.0 support.

krzyzanowskim:
👍 

nonetherless... I'd rather wait for first beta of Swift 3.0 because language evolve so much it's hard to keep the pace now.

krzyzanowskim:
anyway feel free PR to "swift3" branch when ready.

colemancda:
Thanks!

	Coleman,

colemancda:
@krzyzanowskim PR #241 is ready for merge.

colemancda:
@krzyzanowskim PR #244 is a cleaner version of #241 (only 3 commits). 
Also got the unit tests to build and run on Linux.

kinosang:
Great.

I'm a newbie to Swift and I'm working on Mac OS X with Trunk Development Version of swift. How can I specify the branch in `Package.swift`?

krzyzanowskim:
Swift 3 codebase is available from [**swift3**](https://github.com/krzyzanowskim/CryptoSwift/tree/swift3) branch.

This is highly preliminary support for Swift 3 and everything may change.

kdawgwilk:
The swift3 branch is not currently compiling with Xcode 8 beta 2.  I have tried converting it for a PR but I have no idea what I am doing.  I will keep trying but maybe someone wouldn't mind lending a hand

krzyzanowskim:
@kdawgwilk well... what problem you notice? I don't see any problem. It build clean here. I updated few warning after beta 2 releases too.

kdawgwilk:
Well I was right about one thing, I have no idea what I am doing haha I was trying to use the swift3 branch with SPM and couldn't get it to pull the right code.  I ended up adding a tag `0.0.0` locally to the repo and then using this line in my Package.swift
``` swift
.Package(url: "../../CryptoSwift", Version(0,0,0))
```

enagorny:
Does anyone manage to install it via Carthage?

zaubara:
Nope, Carthage build fails:
[...]

PR #302, opened 2016-08-11
Solved bitmap error.

uny:
I recreated the project file with Xcode 8 beta 5, which solved the issue https://github.com/krzyzanowskim/CryptoSwift/pull/282 https://github.com/krzyzanowskim/CryptoSwift/issues/240#issuecomment-233434171.

There are differences of optimization level and they might cause the error.
If you don't care about the differences, the issue will be solved by merging this PR.


krzyzanowskim:
Oh Christ, are you able to separate if that's the optimization level that cause the problem ?

krzyzanowskim:
@uny I wont merge this PR (breaks SPM, change target names), although it's great insight to Xcode 8 issues. Wonder if we can separate what exactly cause the problem to Xcode 8

uny:
Sure, I did.
Link-time optiomazation seems to be the cause.

krzyzanowskim:
From what I found it's not the parameter, I mean by that that maybe it fixes something by accident but in general it's not the ultimate workaround. Still waiting for next betas.

PR #308, opened 2016-08-21
Bugfix and optimization.

mxl:
Tests were failing (and my app was crashing) if `result.count` is greater then `MemoryLayout<UInt32>.size`.

### DIFF
--- a/Sources/CryptoSwift/Collection+Extension.swift
+++ b/Sources/CryptoSwift/Collection+Extension.swift
@@ -8,17 +8,45 @@
 
 extension Collection where Self.Iterator.Element == UInt8, Self.Index == Int {
     func toUInt32Array() -> Array<UInt32> {
-        return stride(from: self.startIndex, to: self.endIndex, by: MemoryLayout<UInt32>.size).map { i in
-            return UnsafePointer(Array(self[Range(i..<i.advanced(by:MemoryLayout<UInt32>.size))]))
-                .withMemoryRebound(to: UInt32.self, capacity: 1) { $0.pointee }
+        var result = Array<UInt32>()
+        result.reserveCapacity(16)
+        for idx in stride(from: self.startIndex, to: self.endIndex, by: MemoryLayout<UInt32>.size) {
+            var val: UInt32 = 0
+            val |= self.count > 3 ? UInt32(self[idx.advanced(by: 3)]) << 24 : 0
+            val |= self.count > 2 ? UInt32(self[idx.advanced(by: 2)]) << 16 : 0
+            val |= self.count > 1 ? UInt32(self[idx.advanced(by: 1)]) << 8  : 0
+            val |= self.count > 0 ? UInt32(self[idx]) : 0
+            result.append(val)
         }
+
+        for _ in result.count..<MemoryLayout<UInt32>.size {
+            result.append(0)
+        }
+
+        return result
     }
 
     func toUInt64Array() -> Array<UInt64> {
-        return stride(from: self.startIndex, to: self.endIndex, by: MemoryLayout<UInt64>.size).map { i in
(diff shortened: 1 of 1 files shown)


krzyzanowskim:
Great catch. Thank you. The last append I added is complete nonsense. It is enough to remove it. 
Although change you've made is clever, I believe is less performant. map() proved to be slow for most cases than for-in, additionally there is more allocs in map closure (one array per each chunk)

mxl:
@krzyzanowskim I'll run some benchmarks. Actually it can have same performance because each  array is turned into UInt32 through `UnsafePointer.withMemoryRebound()` without copy so the real difference is `map` vs `for..in` and shifts.

mxl:
I made few benchmarks and you're right. My fix is 3-4 times slower:
```
import Foundation

extension Collection where Self.Iterator.Element == UInt8, Self.Index == Int {
    func toUInt32Array() -> Array<UInt32> {
        var result = Array<UInt32>()
        for idx in stride(from: self.startIndex, to: self.endIndex, by: MemoryLayout<UInt32>.size) {
            var val: UInt32 = 0
            val |= self.count > 3 ? UInt32(self[idx.advanced(by: 3)]) << 24 : 0
            val |= self.count > 2 ? UInt32(self[idx.advanced(by: 2)]) << 16 : 0
            val |= self.count > 1 ? UInt32(self[idx.advanced(by: 1)]) << 8  : 0
            val |= self.count > 0 ? UInt32(self[idx]) : 0
            result.append(val)
        }

        return result
    }
    
    func toUInt32Array2() -> Array<UInt32> {
        var result = Array<UInt32>()
        result.reserveCapacity(Int(self.count.toIntMax() / MemoryLayout<UInt32>.size.toIntMax()) + 1)
        for idx in stride(from: self.startIndex, to: self.endIndex, by: MemoryLayout<UInt32>.size) {
            var val: UInt32 = 0
            val |= self.count > 3 ? UInt32(self[idx.advanced(by: 3)]) << 24 : 0
            val |= self.count > 2 ? UInt32(self[idx.advanced(by: 2)]) << 16 : 0
            val |= self.count > 1 ? UInt32(self[idx.advanced(by: 1)]) << 8  : 0
            val |= self.count > 0 ? UInt32(self[idx]) : 0
            result.append(val)
        }

        return result
    }

    func toUInt32Array3() -> Array<UInt32> {
        var result = Array<UInt32>()
        for idx in stride(from: self.startIndex, to: self.endIndex, by: MemoryLayout<UInt32>.size) {
            var val: UInt32 = 0
            for byteIdx in (0..<MemoryLayout<UInt32>.size).reversed() {
                val |= self.count.toIntMax() > byteIdx.toIntMax() ? UInt32(self[idx.advanced(by: byteIdx)]) << UInt32(8*byteIdx) : 0
            }
            result.append(val)
        }
        
        return result
    }
    
    func toUInt32Array4() -> Array<UInt32> {
        return stride(from: self.startIndex, to: self.endIndex, by: MemoryLayout<UInt32>.size).map { i in
            return UnsafePointer(Array(self[Range(i..<i.advanced(by:MemoryLayout<UInt32>.size))]))
                .withMemoryRebound(to: UInt32.self, capacity: 1) { $0.pointee }
        }
    }
    
}

let array = (0..<1048576).map { _ in UInt8(arc4random_uniform(256)) }
var startDate = Date()

func start() {
    startDate = Date()
}

func end() -> TimeInterval {
    return Date().timeIntervalSince(startDate)  
}

start()
_ = array.toUInt32Array()
print(end())

start()
_ = array.toUInt32Array2()
print(end())

start()
_ = array.toUInt32Array3()
print(end())

start()
_ = array.toUInt32Array4()
print(end())
```

results in
```
0.28132301568985
0.280844986438751
0.847709953784943
0.950453042984009
```

So i just removed your last append.

krzyzanowskim:
Thanks for you work here!
I already fixed that earlier, so no need to merge now.

PR #309, opened 2016-08-21
Updated travis-ci integration.

mxl:
- Switched osx_image to `xcode8`.
- Updated `swiftenv-install.sh` script to latest version.
- Copied CI.sh inside repository because it will be simpler to maintain it and adapt to project needs. Also replaced `swift-package` with `swift package`.
- Removed linux from os list because swiftenv does not have 3.0 version in `swiftenv install --list` so build always fails on linux. We have to wait until Swift 3.0 will released to include linux back to os list.


krzyzanowskim:
Linux have 3.0 previews, can't be used?
```sh
$ swiftenv install "3.0-PREVIEW-6"
```

krzyzanowskim:
awesome. Can you please move script to scripts/ directory?

mxl:
Finally build succeeded on both Linux and OS X https://travis-ci.org/mxl/CryptoSwift/builds/154212515

PR #321, opened 2016-09-02
Fix SHA512 and SHA384 padding implementation

### LINKED ISSUE #319: SHA512, SHA384 bug
Hi @krzyzanowskim! First of all, great work on implementing the standard crypto functions in Swift.

I'm also implementing crypto functions in Swift, and I've been comparing your digests' outputs to mine, `shasum`'s and other online generators. I came across this bug in CryptoSwift for SHA2 variants that have 64-bit words (i.e. SHA512), where they produce the wrong output for message lengths that are greater than 112 bytes (mod 128).

Ex.

**In CryptoSwift:**
`"abcdefghbcdefghicdefghijdefghijkefghijklfghijklmghijklmnhijklmnoijklmnopjklmnopqklmnopqrlmnopqrsmnopqrstnopqrstu".sha512()` produces `"feb885c0c56838413625ba4e15f3e6116fdb2ebe26f9c1d4b9ea8bea9a984702a61bbdbad784fa995d9ab8879befab0dd7d8566d26a5ad95cdf42fdddf088123"`

**In Terminal using `shasum`:**
`echo -n "abcdefghbcdefghicdefghijdefghijkefghijklfghijklmghijklmnhijklmnoijklmnopjklmnopqklmnopqrlmnopqrsmnopqrstnopqrstu" | shasum -a 512` produces `"8e959b75dae313da8cf4f72814fc143f8f7779c6eb9f7fa17299aeadb6889018501d289e4900f7e4331b99dec4b5433ac7d329eeb6dd26545e96e55b874be909"`

---

I've looked through your code and other implementations in other languages and found that the message "pre-processing" should be different from the other variants. Example implementation in Go can be found [here](https://golang.org/src/crypto/sha512/sha512.go#L210) line 210 wherein the message length representation should be a 128-bit or 16-byte integer (CryptoSwift uses 64-bit or 8-byte which is located [here](https://github.com/krzyzanowskim/CryptoSwift/blob/swift3/Sources/CryptoSwift/SHA2.swift#L199-L209)). This also makes the padding length equal to 112 (mod 128).

I could modify the code and create a pull request but I'm not sure where to put the custom implementation for `HashProtocol.prepare(_ len: Int) -> Array<UInt8>`.

chrisamanse:
I've also added an additional test for SHA384. Should fix #319

### DIFF
--- a/Sources/CryptoSwift/HashProtocol.swift
+++ b/Sources/CryptoSwift/HashProtocol.swift
@@ -16,6 +16,10 @@ internal protocol HashProtocol: class {
 extension HashProtocol {
     
     func prepare(_ len:Int) -> Array<UInt8> {
+        return self.prepare(len, allowance: len / 8)
+    }
+    
+    func prepare(_ len:Int, allowance: Int) -> Array<UInt8> {
         var tmpMessage = message
         
         // Step 1. Append Padding Bits
@@ -25,11 +29,11 @@ extension HashProtocol {
         var msgLength = tmpMessage.count
         var counter = 0
         
-        while msgLength % len != (len - 8) {
+        while msgLength % len != (len - allowance) {
             counter += 1
             msgLength += 1
         }
-
+        
         tmpMessage += Array<UInt8>(repeating: 0, count: counter)
         return tmpMessage
     }
--- a/Sources/CryptoSwift/SHA2.swift
+++ b/Sources/CryptoSwift/SHA2.swift
@@ -196,7 +196,7 @@ final class SHA2 : HashProtocol {
     }
     
     func calculate64() -> Array<UInt8> {
-        var tmpMessage = self.prepare(128)
+        var tmpMessage = self.prepare(128, allowance: 16)
         
         // hash values
         var hh = Array<UInt64>()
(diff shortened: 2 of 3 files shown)


Sources/CryptoSwift/HashProtocol.swift
```diff
@@ -16,6 +16,10 @@ internal protocol HashProtocol: class {
 extension HashProtocol {
     
     func prepare(_ len:Int) -> Array<UInt8> {
+        return self.prepare(len, allowance: 8)
+    }
+    
+    func prepare(_ len:Int, allowance: Int) -> Array<UInt8> {
```
krzyzanowskim: isn't it that allowance is equal to `len / 8` ?

chrisamanse: I'm not sure. There may be some instances where one might want a 0 (mod 128) padding but with only an 8-byte integer representation of the message length - making the allowance to be only 8 rather than 128 / 8 or 16.

If `len / 8` is mostly used, we can set the default return of `prepare(_ len: Int) -> Array<UInt8>` to `return self.prepare(len, allowance: len / 8)` and retain the `prepare(_:allowance:)` function to have a function for customizing the allowance.

krzyzanowskim: Agreed. Please keep it then, and remove `prepare(len:)`. This way it will be explicit.

speaking of, it may be good idea to rename this function to `bitPadding` (as this is what it does), and move it  to Utils.swift

chrisamanse: What I meant was to simply change `return self.prepare(len, allowance: 8)` to `return self.prepare(len, allowance: len / 8)`.

IMHO, `prepare` is okay for the function name as the RFC implementation calls it `pre-processing`.

krzyzanowskim: it is ok name without specific implementation, that is bit padding now. It don't have to be like this. You've made a right point that it's general protocol and now I'm not sure if it should have shared prepare at all. 

for the purpose of this PR, this is how I see it:

```swift
func prepare(_ len:Int, allowance: Int = 0) -> Array<UInt8> {
}
```

overall, I'm very happy with the fix you provided.

PR #324, opened 2016-09-08
Swift3

vincentsaluzzo:
Before I remove bit code compatibility to have compilation successful in Xcode 8 Beta 6, but with the GM, this morning I can build with bit code enabled so a make a PR because all the branch compile with success


krzyzanowskim:
Closed as duplicate. #325

PR #335, opened 2016-09-17
Added Swift 2.3 support

Sumolari:
This adds support for both Swift 2.2 and 2.3.

I've added Xcode 8 test environment but I've also removed Linux environment as I haven't been able to properly set it up. It should pass tests but some help fixing Linux environment to check that automatically would be great.

Results of new test environments: [Build Status](https://travis-ci.org/Sumolari/CryptoSwift)


krzyzanowskim:
Can you please make it available for linux? the script that uses swiftenv is very usefull for that, maybe there's no need to change that?

Sumolari:
I tried to but I haven't been able to set up a Linux environment running Swift 2.x (hard part is getting swift 2.x to build the project), that's why I removed it from Travis settings.

It seems that current [swift2 branch doesn't pass tests on Linux](https://travis-ci.org/krzyzanowskim/CryptoSwift/builds/159454409), too, so maybe we could just ignore Linux tests in this branch, considering this pull request might be the last contribution to it (unless you have plans to continue supporting Swift 2.x on future releases).

PR #353, opened 2016-10-06
Added Crossplatform Randomizer

jeffreybergier:
Hi,

I'm trying to write a simple piece of code that AES encrypts a passcode with a key and IV. The key and IV are generated randomly every time. This code works fine on macOS. But on Linux, the random keys are the same every time the program runs. I have had this problem before. The basic random functions on linux are Turnstile in nature. I found a solution with my own code. It was a basic class that opened /dev/random and read bits off of it on demand. I was wondering if you would be interested in a PR for Multiplatform.swift where the linux aspect of randomization was replaced with this randomizer I've been using on linux.

The code is originally from https://github.com/stormpath/Turnstile/blob/master/Sources/TurnstileCrypto/URandom.swift 

After creating this fork and changing out the randomizer I realized that Crypto swift doesn't actually use the randomizer. So let me know if you actually want to go in the direction of removing the randomizer instead of improving it :p. Or if you have anything you want me to change.

Thanks!

-Jeff

### DIFF
--- a/Sources/CryptoSwift/Cryptors.swift
+++ b/Sources/CryptoSwift/Cryptors.swift
@@ -30,10 +30,6 @@ public protocol Cryptors: class {
 
 extension Cryptors {
     static public func randomIV(_ blockSize:Int) -> Array<UInt8> {
-        var randomIV:Array<UInt8> = Array<UInt8>();
-        for _ in 0..<blockSize {
-            randomIV.append(UInt8(truncatingBitPattern: cs_arc4random_uniform(256)));
-        }
-        return randomIV
+        return URandom.shared.random(numBytes: blockSize)
     }
 }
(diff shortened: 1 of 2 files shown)


krzyzanowskim:
thanks. though, please dont copy code from another project to this project if you're not the author. this is due to licencing.

I dont like singletone solution too :-(

Because of that I wont merge this change. There is nicest way to have randomizer.

jeffreybergier:
I understand. The other project is apache so it shouldn't be a huge conflict. But I understand.

Also, now that I've actually gone through the process of doing this I feel comfortable that CryptoSwift doesn't use the bad linux randomizer. So I can just use the randomizer in my own project to input random values in AES when using it.

Thanks for taking a look and thanks for the project. Its really useful!

krzyzanowskim:
sure :-)

Its not yet decided if I need randomizer at all. Today its more helper (deprecated) than anything usefull, but Ill need it soon for the stuff Im building, then Ill come with some urandom solution for linux.

PR #357, opened 2016-10-10
Specify tvos and watchos deployment targets

hamchapman:
Not having these specified means builds of projects that use these platforms and have CryptoSwift as a dependency (and use Carthage to manage that dependency) will get errors like the one described here: https://github.com/kylef/JSONWebToken.swift/issues/47

### DIFF
--- a/CryptoSwift.xcodeproj/project.pbxproj
+++ b/CryptoSwift.xcodeproj/project.pbxproj
@@ -707,8 +707,10 @@
 				SWIFT_OPTIMIZATION_LEVEL = "-Onone";
 				SWIFT_VERSION = 3.0;
 				TARGETED_DEVICE_FAMILY = "1,2,3,4";
+				TVOS_DEPLOYMENT_TARGET = 9.0;
 				VERSIONING_SYSTEM = "apple-generic";
 				VERSION_INFO_PREFIX = "";
+				WATCHOS_DEPLOYMENT_TARGET = 2.0;
 			};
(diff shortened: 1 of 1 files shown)


krzyzanowskim:
shouldn't be a problem. Current multiplatform setup is described here:  http://promisekit.org/news/2016/08/Multiplatform-Single-Scheme-Xcode-Projects/
I wouldn't be surprised if this is kind-of-carthage-problem

hamchapman:
It probably could be fixed in Carthage if it always assumed the lowest possible deployment targets for all platforms, but I'm not sure of the viability of that. For now I think this is the simplest (and best) solution.

krzyzanowskim:
I'm not going to back for per target configuration, this shouldn't be a problem. Please compare with PromiseKit, check if you can reproduce this issue with PromiseKit, it should behave the same.

hamchapman:
No that's not what this is. I've followed the description in that Promisekit blog post for the PusherSwift lib so I'm familiar with it.

This is just setting build settings on the (single) CryptoSwift target so that Carthage knows what deployment targets are valid for the various platforms.

krzyzanowskim:
o, I'm sorry, I misread the change, should check diff first.

hamchapman:
No worries - I think I misunderstood your first reply as well! 👍

emaloney:
I was having the same issue. Thanks for merging that in so fast, @krzyzanowskim !

PR #377, opened 2016-12-13
Remove .swift-version file

hamchapman:
This shouldn't be part of the repo as it forces a specific version of Swift when Carthage tries to build the framework (if the user installing the library also has swiftenv installed) when their Swift version is different. For example, Xcode 8.2 is using 3.0.2 now, not 3.0.1.


krzyzanowskim:
This is needed for CI and Linux, and Cocoapods now. 
Swift has a long story of backward incompatibility, and if a version is not specified, the result is unpredictable - mostly cause even more problems.

I'd love to just put "3.0" there and it would work with 3.0.x versions, but I think it's not supported, is it @kylef ?

kylef:
> This shouldn't be part of the repo as it forces a specific version of Swift when Carthage tries to build the framework (if the user installing the library also has swiftenv installed) when their Swift version is different.

I don't think this is correct, Carthage just builds the Xcode project. Xcode has its own mechanism to lock Swift versions. The `SWIFT_VERSION` build flag is set on the project so Xcode is using the value of that (https://github.com/krzyzanowskim/CryptoSwift/blob/94f3e347e7bfaed24115d16b205ede4eb1a81633/CryptoSwift.xcodeproj/project.pbxproj#L732).

Removing the `.swift-version` file will not make any difference to building from Carthage/Xcode.

hamchapman:
Which is what I would have thought too, but when installing https://github.com/kylef/JSONWebToken.swift using Carthage (master branch) and with a `.swift-version` in my repo that said 3.0.2 (after Xcode 8.2 was released) then `carthage update` would fail because no legacy swift version had been set in the project. Removing the `.swift-version` file from CryptoSwift made it all work, hence the PR. 

Maybe there's something else at play though!

hamchapman:
Maybe setting the `SWIFT_VERSION` project setting to just `3.0` would work?

I seem to have that set in a lib here, where I didn't see the same error: https://github.com/pusher/pusher-websocket-swift/blob/master/PusherSwift/PusherSwift.xcodeproj/project.pbxproj#L541

krzyzanowskim:
...it is 3.0 already.

Xcode 8.0 - 3.0
Xcode 8.1 - 3.0.1
Xcode 8.2 - 3.0 again

hamchapman:
I was just going with what Kyle had linked to: https://github.com/krzyzanowskim/CryptoSwift/blob/94f3e347e7bfaed24115d16b205ede4eb1a81633/CryptoSwift.xcodeproj/project.pbxproj#L732

krzyzanowskim:
I see... it's adjusted here https://github.com/krzyzanowskim/CryptoSwift/commit/94f3e347e7bfaed24115d16b205ede4eb1a81633

I'm closing this one.

PR #406, opened 2017-02-20
Faster Array(hex:)

nextgenappsllc:
I was refactoring an old extension I made to convert hex strings to a byte array. After, I ran a comparison between the Array initializer for hex strings and the string extension and the string extension was 4x faster. I included the two tests in the extensions test.


krzyzanowskim:
This is awesome! Would you mind replace Array initializer code with this one? I think the initializer is enough. I did a quick test, and it confirms the performance gain there as well.

Please mind the style to match the rest of the code, like [] -> Array<UInt8> and split the code between lines please.

nextgenappsllc:
Sure no problem, I'll make those changes now. Thanks!

nextgenappsllc:
I kept the String extension hexToBytes() in addition to the Array initializer because the string extension is still almost 2x as fast despite using the same code. Check it out and let me know

nextgenappsllc:
Nice it finally passed the error it was getting from my use of NSError. The String.streamHexBytes method that now backs the string extension and array initializer could also be useful for large sequences you don't want to load into an array. An alternative to the unicodeHexMap that works nicely as well in the function is: 

>            guard skip == 0 else {
>                 skip -= 1
>                 continue
>             }
>             guard char.value <= 255 else {return []}
>             var v:UInt8?
>             let c = UInt8(char.value)
>             switch c{
>             case 48...57:
>                 v = c - 48
>             case 65...70:
>                 v = c - 55
>             case 97...102:
>                 v = c - 87
>             default:
>                 break
>             }
>             guard let value = v else {return []}

Although in the new code is would be a throw instead of return.

krzyzanowskim:
Great job. Let's try not to overengineer this helper. Let's go with the `Array(hex:)` only. From my tests, it is significantly faster than the current version.

Keep the `Directory<A,B>` and `Array<>` style. Use `reserveCapacity` for the buffers. and does not make public what don't have to be public.

This is what I tested

```swift
    public init(hex2: String) {
        self.init()
        let unicodeHexMap:[UnicodeScalar:UInt8] = ["0":0,"1":1,"2":2,"3":3,"4":4,"5":5,"6":6,"7":7,"8":8,"9":9,"a":10,"b":11,"c":12,"d":13,"e":14,"f":15]
        var bytes = Array<UInt8>(reserveCapacity: hex2.utf8.underestimatedCount)
        var buffer: UInt8? = 0
        var skip = hex2.hasPrefix("0x") ? 2 : 0
        for char in hex2.unicodeScalars.lazy {
            guard skip == 0 else {
                skip -= 1;
                continue
            }
            guard let value = unicodeHexMap[char] else {
                continue
            }
            if let b = buffer {
                let byte = b << 4 | value
                bytes.append(byte)
                buffer = nil
            } else {
                buffer = value
            }
        }
        if let b = buffer {
            self.append(b as! Element)
        }
    }
```

The version with `case` switch looks better, though, not sure how about the performance. I think I had such code at some point, then I moved to the Stdlib... but it just confirms it's rarely a good idea.

nextgenappsllc:
Yeah I felt it getting a bit over done, I just wanted to make sure the Array initializer was using the same exact code as the string extension because I was trying to figure out why the string extension still performed better after making the changes in the Array initializer (Which I think is because the cast to element). 

I'll push a new build with the changes. The unicodeHexMap in your example and in my original pull request forgot about uppercase letters having different codes so I will add those into it although, I think the case statement performed the same so I'll confirm and include it if so.

nextgenappsllc:
I tested the switch vs dictionary performance and the dictionary did better so I left it. But then I got curious as to why and removed the ranges which changed the switch to: 

 

>            guard char.value <= 255 else {
>                 self.removeAll()
>                 return
>             }
>             var v:UInt8?
>             let c:UInt8 = UInt8(char.value)
>             switch c{
>             case 48:
>                 v = 0
>             case 49:
>                 v = 1
>             case 50:
>                 v = 2
>             case 51:
>                 v = 3
>             case 52:
>                 v = 4
>             case 53:
>                 v = 5
>             case 54:
>                 v = 6
>             case 55:
>                 v = 7
>             case 56:
>                 v = 8
>             case 57:
>                 v = 9
>             case 65,97:
>                 v = 10
>             case 66,98:
>                 v = 11
>             case 67,99:
>                 v = 12
>             case 68,100:
>                 v = 13
>             case 69,101:
>                 v = 14
>             case 70,102:
>                 v = 15
>             default:
>                 break
>             }
>             guard let value = v else {
>                 self.removeAll()
>                 return
>             }


And it performed better than the dictionary oddly enough.

krzyzanowskim:
We can stay with the dictionary if it's significantly better, nothing wrong with it. Ok, is it ready to merge now?

nextgenappsllc:
I refactored the switch to condense it without using ranges and it performs actually better than the dictionary and included it in the last build. Should be all set to merge!

PR #441, opened 2017-06-06
Compile under Xcode 9 / Swift 3.2

Dimillian:
This is some very lite change I've done to make CryptoSwift compile under the new Xcode and Swift 3.2 compiler.

### DIFF
--- a/Sources/CryptoSwift/AES.swift
+++ b/Sources/CryptoSwift/AES.swift
@@ -197,12 +197,27 @@ fileprivate extension AES {
         b2 = F1(t[2], t[3], t[0], t[1]) ^ rk[rounds][2]
         b3 = F1(t[3], t[0], t[1], t[2]) ^ rk[rounds][3]
 
+        let r0 = UInt8(b0 & 0xFF)
+        let r1 = UInt8((b0 >> 8) & 0xFF)
+        let r2 = UInt8((b0 >> 16) & 0xFF)
+        let r3 = UInt8((b0 >> 24) & 0xFF)
+        let r4 = UInt8(b1 & 0xFF)
+        let r5 = UInt8((b1 >> 8) & 0xFF)
+        let r6 = UInt8((b1 >> 16) & 0xFF)
+        let r7 = UInt8((b1 >> 24) & 0xFF)
+        let r8 = UInt8(b2 & 0xFF)
+        let r9 = UInt8((b2 >> 8) & 0xFF)
+        let r10 = UInt8((b2 >> 16) & 0xFF)
+        let r11 = UInt8((b2 >> 24) & 0xFF)
+        let r12 = UInt8(b3 & 0xFF)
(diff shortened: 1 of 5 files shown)


krzyzanowskim:
Thank you! awesome. Don't get it wrong, I won't merge the changes. I already started to work on it by myself and I hope not all changes are needed though

Dimillian:
Of course, this just for me, need my project to compile ;)

Sources/CryptoSwift/AES.swift
```diff
@@ -197,12 +197,27 @@ fileprivate extension AES {
...
+        let r8 = UInt8(b2 & 0xFF)
+        let r9 = UInt8((b2 >> 8) & 0xFF)
+        let r10 = UInt8((b2 >> 16) & 0xFF)
+        let r11 = UInt8((b2 >> 24) & 0xFF)
+        let r12 = UInt8(b3 & 0xFF)
+        let r13 = UInt8((b3 >> 8) & 0xFF)
+        let r14 = UInt8((b3 >> 16) & 0xFF)
+        let r15 = UInt8((b3 >> 24) & 0xFF)
         return [
-            UInt8(b0 & 0xFF),UInt8((b0 >> 8) & 0xFF),UInt8((b0 >> 16) & 0xFF),UInt8((b0 >> 24) & 0xFF),
```
krzyzanowskim: hm... this shouldn't be needed. why is it so? because "it's too complicated to blabla" ?

gonghao: Yes. I think it's for the "too complicated to blabla".

Dimillian: Yeah, it may be a compiler bug in B1, would be good to report it to Apple.

krzyzanowskim:
Swift 3.2 and Xcode 9 compatibility is available from the `develop` branch

SpacyRicochet:
No chance of making this an actual release, is there?

We actually use CryptoSwift as part of a private pod, but podspecs don't like you pointing them towards `:git` URLs and branches.

krzyzanowskim:
hm.... I could do beta release maybe. Hold my 🍺

SpacyRicochet:
Did you get a chance to make a beta release on the `develop` branch? Would be awesome.

Meanwhile, we've forked it to manage our own Swift 3.2 compatibility, but `pod lib lint` is complaining about frameworks or tons of build errors. Very confusing.

krzyzanowskim:
yes. Can't push to CocoaPods though due to backend error 500 :-( reported to CocoaPods

SpacyRicochet:
Ah, that's too bad. I wanted to let you know that we got it working through a fork and a private pod repo now. So we're good now.

The release is probably still great for other people, so I can only encourage you to continue trying :)

And thanks for looking into this for us! ❤️

krzyzanowskim:
🚀  CryptoSwift (0.7.0-alpha.1) successfully published

PR #442, opened 2017-06-06
Support Swift4 on Xcode9 beta

levantAJ:


krzyzanowskim:
Thanks!

It not really Swift 4, more like Swift 3.2 isn't it? Anyway, I pushed some of the changes to the `develop` branch to support Swift 3.2. There is still "the expression was too complex...." regression https://bugs.swift.org/browse/SR-5140 but I'd rather to workaround that than change the code therefore it's not yet changed on the branch.

PR #445, opened 2017-06-07
Fix compile error

ldrr:
Simplify command by predefine return variable type


krzyzanowskim:
love this workaround ;) will apply it upstream with the other changes though. great find @ldrr

PR #463, opened 2017-07-26
Make Carthage cache work

beseder42:
## What's in this pull request?
Update the project configurations for generating CryptoSwift-Swift.h to let Carthage know Swift version by which the framework is built.

## New folder structure of the built framework
folderstructure of product

## Explanation and origin
#https://github.com/typelift/SwiftCheck/pull/227


beseder42:
Errors are because of Swift 4.0, which is not handled atm by Travis.

Issue #447, opened 2017-06-21
CryptoSwift not compiling due to a complex line of code


Joannis:
[This line of code in MD5](https://github.com/krzyzanowskim/CryptoSwift/blob/develop/Sources/CryptoSwift/MD5.swift#L104) is not compiling on my server. It's a 32-core nehalem CPU, and although it's powerful, one single core compiling the MD5.swift file is having trouble with Line 104. I see you already simplified the complex statement a little bit, because you split off one byte from the 32-bit unsigned integer. But I don't see why the others aren't split off, too.

krzyzanowskim:
"to complex in reasonable time" factor is machine dependent. The best you can do is give it more time to compile https://github.com/krzyzanowskim/CryptoSwift/blob/master/scripts/CI.sh#L26

krzyzanowskim:
see https://github.com/krzyzanowskim/CryptoSwift/issues/418#issuecomment-291589701

Joannis:
I'm aware, thanks for the reminder of the workaround :)
But I don't see the added value of keeping them on a single line of code there. It prevents compilation on older machines (such as any Nehalem CPU), which is a downside I can see.

I'll use the workaround for now anyways :)

Issue #418, opened 2017-03-27
Cannot compile with Jenkins using 0.6.8


zeusent:
Since updating to version `0.6.8` compilation via Jenkins fails with these errors in `AES.swift`:

```
/Users/jenkins/Jenkins/jobs/[MY_PROJECT]/workspace/Pods/CryptoSwift/Sources/CryptoSwift/AES.swift:142:179: error: expression was too complex to be solved in reasonable time; consider breaking up the expression into distinct sub-expressions
        var b0 = UInt32(block[block.startIndex + 0 + (0 << 2)]) << 0 | UInt32(block[block.startIndex + 1 + (0 << 2)]) << 8 | UInt32(block[block.startIndex + 2 + (0 << 2)]) << 16 | UInt32(block[block.startIndex + 3 + (0 << 2)]) << 24
                 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/Users/jenkins/Jenkins/jobs/[MY_PROJECT]/workspace/Pods/CryptoSwift/Sources/CryptoSwift/AES.swift:143:179: error: expression was too complex to be solved in reasonable time; consider breaking up the expression into distinct sub-expressions
        var b1 = UInt32(block[block.startIndex + 0 + (1 << 2)]) << 0 | UInt32(block[block.startIndex + 1 + (1 << 2)]) << 8 | UInt32(block[block.startIndex + 2 + (1 << 2)]) << 16 | UInt32(block[block.startIndex + 3 + (1 << 2)]) << 24
[...]

PR #467, opened 2017-08-07
Reduce expression complexity to appease compiler [Xcode 9]

adolfo:
This PR addresses a new issue in Xcode 9 Beta 5 (as well as existing issues #447, #418) where the compiler is unable to resolve certain expressions within a reasonable amount of time. The solution here, as before, is to break down these one-liners into multiple, simpler expressions.

### DIFF
--- a/Sources/CryptoSwift/AES.swift
+++ b/Sources/CryptoSwift/AES.swift
@@ -307,19 +307,14 @@ fileprivate extension AES {
         var rk2: Array<Array<UInt32>> = expandKey(key, variant: variant)
 
         for r in 1 ..< rounds {
-            var w: UInt32
-
-            w = rk2[r][0]
-            rk2[r][0] = U1[Int(B0(w))] ^ U2[Int(B1(w))] ^ U3[Int(B2(w))] ^ U4[Int(B3(w))]
-
-            w = rk2[r][1]
-            rk2[r][1] = U1[Int(B0(w))] ^ U2[Int(B1(w))] ^ U3[Int(B2(w))] ^ U4[Int(B3(w))]
-
-            w = rk2[r][2]
-            rk2[r][2] = U1[Int(B0(w))] ^ U2[Int(B1(w))] ^ U3[Int(B2(w))] ^ U4[Int(B3(w))]
-
-            w = rk2[r][3]
-            rk2[r][3] = U1[Int(B0(w))] ^ U2[Int(B1(w))] ^ U3[Int(B2(w))] ^ U4[Int(B3(w))]
+            for i in 0..<4 {
+                let w = rk2[r][i]
+                let u1 = U1[Int(B0(w))]
+                let u2 = U2[Int(B1(w))]
+                let u3 = U3[Int(B2(w))]
+                let u4 = U4[Int(B3(w))]
+                rk2[r][i] = u1^u2^u3^u4
+            }
         }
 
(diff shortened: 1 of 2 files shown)


Sources/CryptoSwift/MD5.swift
```diff
@@ -101,10 +101,14 @@ public final class MD5: DigestType {
...
             let gAdvanced = g << 2
-            var Mg = UInt32(chunk[chunk.startIndex &+ gAdvanced]) | UInt32(chunk[chunk.startIndex &+ gAdvanced &+ 1]) << 8 | UInt32(chunk[chunk.startIndex &+ gAdvanced &+ 2]) << 16
-                Mg = Mg | UInt32(chunk[chunk.startIndex &+ gAdvanced &+ 3]) << 24
 
-            B = B &+ rotateLeft(A &+ F &+ k[j] &+ Mg, by: s[j])
+            let mg0 = UInt32(chunk[chunk.startIndex &+ gAdvanced])
+            let mg1 = UInt32(chunk[chunk.startIndex &+ gAdvanced &+ 1]) << 8
+            let mg2 = UInt32(chunk[chunk.startIndex &+ gAdvanced &+ 2]) << 16
+            let mg3 = UInt32(chunk[chunk.startIndex &+ gAdvanced &+ 3]) << 24
+            let mg = (mg0 | mg1 | mg2) | mg3
```
krzyzanowskim: Nitpick: `Mg` stands for `M[g]`, I'd like to keep it this way ;)

Sources/CryptoSwift/AES.swift
```diff
@@ -307,19 +307,14 @@ fileprivate extension AES {
...
-
-            w = rk2[r][3]
-            rk2[r][3] = U1[Int(B0(w))] ^ U2[Int(B1(w))] ^ U3[Int(B2(w))] ^ U4[Int(B3(w))]
+            for i in 0..<4 {
+                let w = rk2[r][i]
+                let u1 = U1[Int(B0(w))]
+                let u2 = U2[Int(B1(w))]
+                let u3 = U3[Int(B2(w))]
+                let u4 = U4[Int(B3(w))]
+                rk2[r][i] = u1^u2^u3^u4
```
krzyzanowskim: I wonder how it impacts the performance, the rough access is more likely to be optimized. Did you test the performance diff for that change?

adolfo: Other than running the unit tests I didn't run any performance tests. I think we gained something by removing that `var` but I do see now how that `0..<4` loop could affect performance if the compiler isn't smart enough to optimize a "constant" range. I can break those out into 4 copies.

Sources/CryptoSwift/MD5.swift
adolfo: 👍🏼

Sources/CryptoSwift/AES.swift
adolfo: I tested this using the performance unit tests and here's what I found.
Using for loop vs duplicate expressions:
encrypt: 2.313 vs 2.302 seconds
decrypt: 3.712 vs 3.806 seconds

adolfo: ...and as a baseline the performance before any of these changes was:
encrypt: 2.414
decrypt: 3.816

PR #499, opened 2017-09-21
Add support for Ethereum compatible Keccak SHA3

zweigraf:
Hello!

I needed a Keccak256 implementation for an Ethereum project. I found your library and gave it a try. Unfortunately (for me), it uses the standardized SHA3, not the original Keccak one which is used in Ethereum/Solidity projects. As I found out, the implementations only differ in `0x06` vs `0x01` (I don't know the correct name for this, so I called it `markByte`, gathered from context). Changing this works to produce the desired results. 

Adding the necessary variants incurred some problems regarding the RawRepresentability of the enum, but I could not see any use of this in the library itself, that's why I deleted it. If you have any concerns about compatibility, let me know.

Expected test results are taken from <https://emn178.github.io/online-tools/keccak_256.html> (yes, even the one million 'a' ones). 

Please tell me if you are interested in this change and if I should change some stuff (e.g. create another separate `DigestType` for Keccak).


zweigraf:
Glad to hear! Be aware though: removing the RawRepresentability could be considered a major change I guess.

krzyzanowskim:
yea. would make sense to remove it from the SHA2.Variant as well.

zweigraf:
Yay I can delete my fork now :)

ashchan:
Wonderful, now I got rid of another C lib. Thank you guys!

PR #556, opened 2018-01-02
Remove deprecated 'characters'

chrishannah:
Change string length to use utf8 length, rather than the deprecated characters variable.


krzyzanowskim:
the branch `swift4` is obsolete, only for Xcode 8. With Xcode 9 you should use master branch

PR #567, opened 2018-01-18
AEAD using ChaCha20-Poly1305

spatno:
As both ChaCha20 and Poly1305 are often used in combination as described in [rfc7539](https://tools.ietf.org/html/rfc7539), I've created some simple functions to utilize them together in that fashion.


Sources/CryptoSwift/ChaCha20Poly1305.swift
```diff
@@ -0,0 +1,83 @@
...
+//  Permission is granted to anyone to use this software for any purpose,including commercial applications, and to alter it and redistribute it freely, subject to the following restrictions:
+//
+//  - The origin of this software must not be misrepresented; you must not claim that you wrote the original software. If you use this software in a product, an acknowledgment in the product documentation is required.
+//  - Altered source versions must be plainly marked as such, and must not be misrepresented as being the original software.
+//  - This notice may not be removed or altered from any source or binary distribution.
+//
+//
+//  https://tools.ietf.org/html/rfc7539#section-2.8.1
+
+import UIKit
```
krzyzanowskim: This is probably not needed at all.
See other conditional imports.

Sources/CryptoSwift/ChaCha20Poly1305.swift
```diff
@@ -0,0 +1,83 @@
...
+    public class func decrypt( cipher:Array<UInt8>, header:Array<UInt8>, key:Array<UInt8>, nonce:Array<UInt8>, tag:Array<UInt8> ) throws -> (message:Array<UInt8>, success:Bool) {
+        
+        let chacha = try ChaCha20(key: key, iv: nonce)
+        
+        var polykey = Array<UInt8>(repeating: 0, count: 32)
+        polykey = try chacha.encrypt( polykey )
+        
+        let mac = try ChaCha20Poly1305.calculateTag(cipher: cipher, header: header, encodedKey: polykey)
+        
+        if mac == tag {
```
krzyzanowskim: make it `guard` and return early.

krzyzanowskim: instead of empty array, better return input (header).

Sources/CryptoSwift/ChaCha20Poly1305.swift
```diff
@@ -0,0 +1,83 @@
...
+            return (message, true)
+        } else {
+            return ([], false)
+        }
+    }
+    
+    private class func calculateTag( cipher:Array<UInt8>, header:Array<UInt8>, encodedKey:Array<UInt8> ) throws -> Array<UInt8> {
+        
+        let poly1305 = Poly1305(key: encodedKey)
+        var mac = Array<UInt8>()
```
krzyzanowskim: since the size is known upfront, use the `Array<UInt8>(reserveCapacity: X)` for better performance

Sources/CryptoSwift/ChaCha20Poly1305.swift
```diff
@@ -0,0 +1,83 @@
...
+        mac.append(contentsOf: header)
+        let headerPadding = ((16 - (header.count & 0xf)) & 0xf)
+        mac.append(contentsOf: Array<UInt8>(repeating: 0, count: headerPadding))
```
krzyzanowskim:
```swift
mac += Array<UInt8>(repeating: 0, count: headerPadding)
```

Sources/CryptoSwift/ChaCha20Poly1305.swift
```diff
@@ -0,0 +1,83 @@
...
+        mac.append(contentsOf: cipher)
+        let cipherPadding = ((16 - (cipher.count & 0xf)) & 0xf)
+        mac.append(contentsOf: Array<UInt8>(repeating: 0, count: cipherPadding))
+        let headerCount = UInt64(header.count).bytes().reversed()
```
krzyzanowskim: if the "reversed()" is due to endianness, we should check the endianness first and use it when needed

Sources/CryptoSwift/ChaCha20Poly1305.swift
```diff
@@ -0,0 +1,83 @@
...
+        let chacha = try ChaCha20(key: key, iv: nonce)
+        
+        var polykey = Array<UInt8>(repeating: 0, count: 32)
+        polykey = try chacha.encrypt( polykey )
+        
+        let mac = try ChaCha20Poly1305.calculateTag(cipher: cipher, header: header, encodedKey: polykey)
+        
+        if mac == tag {
+            var toDecrypt = Array<UInt8>()
+            toDecrypt.append(contentsOf: polykey)
```
krzyzanowskim: toDecrypt += polykey

Sources/CryptoSwift/ChaCha20Poly1305.swift
```diff
@@ -0,0 +1,83 @@
...
+        
+        let chacha = try ChaCha20(key: key, iv: nonce)
+        
+        var polykey = Array<UInt8>(repeating: 0, count: 32)
+        polykey = try chacha.encrypt( polykey )
+        
+        let mac = try ChaCha20Poly1305.calculateTag(cipher: cipher, header: header, encodedKey: polykey)
+        
+        if mac == tag {
+            var toDecrypt = Array<UInt8>()
```
krzyzanowskim: Array<UInt8>(reserveCapacity: X)

Tests/CryptoSwiftTests/ChaCha20Poly1305Tests.swift
```diff
@@ -0,0 +1,76 @@
+//
+//  ChaCha20Poly1305Tests.swift
+//  Tests
```
krzyzanowskim: Adjust header

Tests/CryptoSwiftTests/ChaCha20Poly1305Tests.swift
```diff
@@ -0,0 +1,76 @@
...
+            XCTAssert(false, "Encryption Failed")
+        }
+        
+        do {
+            let decryptResult = try ChaCha20Poly1305.decrypt(cipher: cipherArr, header: headerArr, key: keyArr, nonce: nonceArr, tag: tagArr)
+            
+            XCTAssertEqual(decryptResult.success, true,  "decrypt mac check failed")
+            XCTAssertEqual(decryptResult.message, messageArr, "message not equal")
+        } catch {
+            XCTAssert(false, "Encryption Failed")
```
krzyzanowskim: even if I use it differently, use of `error.localizedDescription` here would be nice to have

Sources/CryptoSwift/ChaCha20Poly1305.swift
```diff
@@ -0,0 +1,83 @@
...
+
+    public class func encrypt( message:Array<UInt8>, header:Array<UInt8>, key:Array<UInt8>, nonce:Array<UInt8> ) throws -> (cipher:Array<UInt8>, tag:Array<UInt8>) {
+        
+        let chacha = try ChaCha20(key: key, iv: nonce)
+        
+        var polykey = Array<UInt8>(repeating: 0, count: 32)
+        var toEncrypt = Array<UInt8>()
+        
+        toEncrypt.append(contentsOf: polykey)
+        polykey = try chacha.encrypt( polykey )
```
krzyzanowskim: what with that spaces :) ? `( polykey )`. Use `SwiftFormat` to format format automatically.

Tests/CryptoSwiftTests/ChaCha20Poly1305Tests.swift
```diff
@@ -0,0 +1,76 @@
...
+        executeTestCase("6b65792e6b65792e6b65792e6b65792e6b65792e6b65792e6b65792e6b65792e",
+               "6e6f6e63652e6e6f6e63652e",
+               "",
+               "6d657373616765",
+               "5d9c0a9fe7d5e5",
+               "c93aa61fc3cc66a819ac96f6ce365aee")
+    }
+    
+    func test2() {
+        /* Test vector from section 2.8.2. */
```
krzyzanowskim: Please reference the source document too.

Sources/CryptoSwift/ChaCha20Poly1305.swift
spatno: I'll use the endianness init of UInt64 to make this more clear

spatno:
Thanks for looking it over! I agree with all the minor code changes. As for making this an extension on ChaCha20, I think a better route may to to introduce a new protocol for AEAD. AEAD is fundamentally a combination of your Authenticator and Cipher protocols. ChaCha20-Poly1305 is one example, but it wouldn't be out of the question for someone to use Salsa20 or AES instead of ChaCha20. 

Something like this:

`public protocol AEAD {
    public class func encrypt( message:Array<UInt8>, header:Array<UInt8>, key:Array<UInt8>, nonce:Array<UInt8> ) throws -> (cipher:Array<UInt8>, tag:Array<UInt8>) 
    
    public class func decrypt( cipher:Array<UInt8>, header:Array<UInt8>, key:Array<UInt8>, nonce:Array<UInt8>, tag:Array<UInt8> ) throws -> (message:Array<UInt8>, success:Bool)
}`

krzyzanowskim:
Go with the AEAD protocol. Sounds good.

krzyzanowskim:
Let me know when you finish work on your side and I should re-review. thanks!

spatno:
Thanks, it's ready for review

krzyzanowskim:
I'm thinking. This could be abstracted out, couldn't it? ChaCha20 is a `Cipher` type, Poly1305 is the `Authenticator`, would it work with any of that automatically? The AEAD could be implemented completely as a protocol then. WDYT?

spatno:
If it can be, that'd certainly be nice, did you have a structure in mind? I'm not sure if it's possible that a generalized Cipher or Authenticator could be passed in to AEAD as I'm not familiar enough with other implementations, but I could do some research. It may be that the interaction of the Cipher and Authenticator, in this particular case Poly1305 and ChaCha20 is different from say Poly1305 and AES. Additionally, there are different approaches to AEAD like doing Encrypt-then-MAC or MAC-then-Encrypt ([wiki](https://en.wikipedia.org/wiki/Authenticated_encryption))

krzyzanowskim:
@spatno I'll take over this if you don't mind. I really like what you did here and I'd love to merge it to develop for further work on that.

spatno:
@krzyzanowskim  Yea no problem, go for it! Just want anyone who would find it useful to have it and extend.

PR #580, opened 2018-02-21
Fix #579

### LINKED ISSUE #579: Big inputs for SHA3 (keccak256) hashes lead to fatal crashes
# Details

- CryptoSwift version: 0.8.2
- Swift version: 4.0.3
- Platform version: 
	- [x] iOS/tvOS/watchOS
	- [x] macOS
	- [ ] Linux
- Installation: 
	- [x] CocoaPods
	- [ ] Carthage
	- [x] Swift Package Manager
	- [ ] Manual
- [ ] [I'm a supporter](https://www.paypal.com/cgi-bin/webscr?cmd=_s-xclick&hosted_button_id=92Z6U3LBHF9J4).

# Description

Generating a SHA3 (keccak256) Hash of a big Byte Array leads sometimes to a crash. It somehow depends on the number of Bytes in the input Array but I don't really know the SHA3 specifications so I can't really help with the solution.

The crash happens [here](https://github.com/krzyzanowskim/CryptoSwift/blob/master/Sources/CryptoSwift/SHA3.swift#L264) and it seems like the `markByteIndex` is out of range for the `accumulated` array, which makes sense as `markByteIndex` is set to `accumulated.count` and in this specific case `accumulated` doesn't get padded as `accumulated.count == 0 || accumulated.count % blockSize != 0` is false.

I don't know how this implementation should look like but this special case should be considered. The crash is reproducible and happens for many different input sizes as selecting it randomly leads to a crash after a few seconds.

screen shot 2018-02-19 at 02 17 01

screen shot 2018-02-19 at 02 17 33

# Steps to reproduce

Sample code:

The following code crashes every single time. You just have to wait a few seconds as random values are generated.

```swift
import CryptoSwift
#if os(Linux) || os(FreeBSD)
    import Glibc
#else
    import Darwin
#endif

// ======== Helpers ========

extension Array where Element == UInt8 {

    static func random(count: Int) -> [UInt8]? {
        var array = [UInt8](repeating: 0, count: count)

        let fd = open("/dev/urandom", O_RDONLY)
        guard fd != -1 else {
            return nil
        }

        let ret = read(fd, &array, MemoryLayout<UInt8>.size * array.count)
        guard ret > 0 else {
            return nil
        }

        return array
    }

    public var bigEndianUInt: UInt? {
        guard self.count <= MemoryLayout<UInt>.size else {
            return nil
        }
        var number: UInt = 0
        for i in (0 ..< self.count).reversed() {
            number = number | (UInt(self[self.count - i - 1]) << (i * 8))
        }

        return number
    }
}

// ======== End Helpers ========

// ======== This code crashes at some point ========

while true {

    guard let randomCount = [UInt8].random(count: 2)?.bigEndianUInt else {
        continue
    }
    guard let randomBytes = [UInt8].random(count: Int(randomCount)) else {
        continue
    }

    let hash = SHA3(variant: .keccak256).calculate(for: randomBytes)
}
```

Ybrin:
SHA3 hashes for multiples of the corresponding block size were
calculated wrongly. This led to a fatal crash.

Now we pad the input even
if the input size is a multiple of the block size as defined in the SHA3
specifications.

Also added some tests for those edge cases.

Fixes #579 

Checklist:
- [x] Tests added.

- Padding for SHA3 input bytes is done correctly now
- Added tests for edge case (multiple of block size) SHA 3 hashes for all variants


krzyzanowskim:
Thanks!
Merged manually by cherry-pick 1a399492c919ca7d267e2d4a62172e0eb7f5d912 to develop

Issue #164, opened 2015-11-17
Linking against dylib not safe for use in application extensions warning


rinatkhanov:
Seeing this warning when installing CryptoSwift to my extension target with CocoaPods:
`linking against dylib not safe for use in application extensions`

Unless I am missing something, looks like it's [pretty easy to fix](https://github.com/ankurp/Dollar.swift/issues/144).

krzyzanowskim:
"Require Only App-Extension-Safe API" is set to "Yes" (except the watchOS target).

PR #585, opened 2018-03-09
Extension safe api set to YES

hamchapman:
Seems like this used to be the case: https://github.com/krzyzanowskim/CryptoSwift/issues/164 but now isn't

Is that for a reason?

Checklist:
- [x] Tests added. (n/a)

- Set `APPLICATION_EXTENSION_API_ONLY = YES`

Otherwise you get this warning when you have it set to `YES` in your own target:

screen shot 2018-03-09 at 14 33 56


.swift-version
```diff
@@ -1 +1 @@
-4.1
\ No newline at end of file
+4.1-DEVELOPMENT-SNAPSHOT-2018-03-08-a
```
krzyzanowskim: no. This supposes to be ready for 4.1 release. I know .swift-version is restrictive, but snapshot will outdate like.... in few days.

hamchapman: Okay cool - I'll remove that change and close the other PR 👍

hamchapman: Updated

krzyzanowskim:
thanks!

Issue #611, opened 2018-04-26
Make variant injectable


zarghol:
# Description

Since old version (0.7.2) could work with a 64-length key, with an aes128 variant,
It could be great to let the user decide which variant to use, for compatibility purpose and for convenience.
We could put as a default the current choice based on the key length.

zarghol:
pull request submitted : https://github.com/krzyzanowskim/CryptoSwift/pull/612

krzyzanowskim:
64 bytes long key is not valid AES key and shouldn't work nor be used ever. As pointed https://github.com/krzyzanowskim/CryptoSwift/pull/612#discussion_r189444431 so far I was not able to reproduce flawless use of 64 bytes long key with 0.7.2

PR #612, opened 2018-04-27
Fix/aes variant

zarghol:
Fixes aes variant

Checklist:
- [x] Tests added.

- AES Variant no longer is a computed var
- AES Variant is initialized at init, with default implementation like before (16 -> 128, 24 -> 192, 32 -> 256) but error is send when the key is not ok for these rules
- AES Variant can be vetted in the init by the user to choose a custom value if needed.


krzyzanowskim:
Tests failing on CI

Test Case 'AESTests.testEncryptTooShortWithVariantKey' started at 2018-04-27 08:37:41.406
Exited with signal code 4

zarghol:
Yes, I don't know why... It works on my computer.

Sources/CryptoSwift/AES.swift
```diff
@@ -48,18 +62,7 @@ public final class AES: BlockCipher {
...
-            return .aes128
-        case 192:
-            return .aes192
-        case 256:
-            return .aes256
-        default:
-            preconditionFailure("Unknown AES variant for given key.")
-        }
-    }
+    public var variant: Variant
```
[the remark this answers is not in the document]
zarghol: yes, but, that way, it let the user change it after init if needed, but since the value is already passed in the init, I can change it if you prefer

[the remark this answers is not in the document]
krzyzanowskim: the value shouldn't be changed at any point later.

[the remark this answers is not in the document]
zarghol: ok, I'll change it so

Tests/Tests/AESTests.swift
```diff
@@ -473,6 +473,55 @@ extension AESTests {
...
+
+    func testEncryptTooShortWithVariantKey() {
+        let key = "f83".bytes
+        let iv = "0000000000000000".bytes
+
+        do {
+            let aes = try AES(key: key, blockMode: CBC(iv: iv), variant: .aes128)
+            _ = try aes.encrypt(key)
+        } catch {
+            XCTFail("an error occured : \(error)")
```
krzyzanowskim: this is negative test, error is expected, test shouldn't fail in this case

zarghol: why is this test should fail ?

krzyzanowskim: the keys is "f83", that should throw error. This PR is all about it, isn't it?

zarghol: The first test tooShort `testEncryptTooShortKey` , is supposed to pass in the catch, so, there is no XCTFailt in there but in the do. 
But the `testEncryptTooShortWithVariantKey ` test use the aes128 variant, so It doesn't use the key length to compute the variant, so it is normal to pass

krzyzanowskim: what is the valid AES variant for key "f83" ?

zarghol: I agree with you that the inferred variant is the invalidKey error, but in my unit test, I am not using the inferred one

Tests/Tests/AESTests.swift
```diff
@@ -507,6 +556,10 @@ extension AESTests {
             ("testAESGCMTestCase5", testAESGCMTestCase5),
             ("testAESGCMTestCase6", testAESGCMTestCase6),
             ("testAESGCMTestCase7", testAESGCMTestCase7),
+            ("testEncryptTooLongKey", testEncryptTooLongKey),
+            ("testEncryptTooLongWithVariantKey", testEncryptTooLongWithVariantKey),
+            ("testEncryptTooShortKey", testEncryptTooShortKey),
+            ("testEncryptTooShortWithVariantKey", testEncryptTooShortWithVariantKey)
```
zarghol: maybe a comma is missing here for Travis CI ?

Sources/CryptoSwift/AES.swift
```diff
@@ -120,11 +133,21 @@ public final class AES: BlockCipher {
...
-    public init(key: Array<UInt8>, blockMode: BlockMode, padding: Padding = .pkcs7) throws {
+    public init(key: Array<UInt8>, blockMode: BlockMode, padding: Padding = .pkcs7, variant: Variant? = nil) throws {
         self.key = Key(bytes: key)
         self.blockMode = blockMode
         self.padding = padding
-        keySize = self.key.count
+        self.keySize = self.key.count
+
+        if let variant = variant {
+            guard variant.keySize <= keySize else {
```
krzyzanowskim: !=

it's not enough to check if it's shorter, is it?

Sources/CryptoSwift/AES.swift
```diff
@@ -120,11 +133,21 @@ public final class AES: BlockCipher {
...
+                throw Error.invalidKeySize
+            }
+            precondition((try? Variant(keySize: key.count)) == variant, "Warning : Key size should be chosen according to the Variant : 16 bytes for AES-128, 24 bytes for AES-192 and 32 bytes for AES-256")
```
krzyzanowskim: the idea was to throw error, not to assert

Sources/CryptoSwift/AES.swift
zarghol: If the size has to be the same, my PR is nonsense : It is intended to use a 64 length key with 128 variant (which has to be a 16 length key based on the current version)

krzyzanowskim: you can't just use 64 bytes when you need 128 bytes. What's the point of that?

zarghol: the 128 variant use a 16 length key. I have a 64 length key, and since it was working on previous versions, I need to keep this key, for compatibility purpose

zarghol: We used the 0.7.2 version

krzyzanowskim: hm... When I'm trying to use 64 bytes length key with AES on 0.7.2 it crashes - that is kinda expected, although the error would be better.

krzyzanowskim:
I'm not going to merge this PR. see https://github.com/krzyzanowskim/CryptoSwift/pull/612#discussion_r189207820

The AES variant validation part is a good idea though, added with this commit:  https://github.com/krzyzanowskim/CryptoSwift/commit/3ff4b5950a355349bc7e10819cec6cdad3cd7205

Thank you for you work! Let's continue discussion in https://github.com/krzyzanowskim/CryptoSwift/issues/611

PR #620, opened 2018-05-19
Fix sample code in README

frankdilo:
Some sample code in the README was referring to `CBC` block mode which now lives in the `BlockMode` enum, so to use it we need to prefix it with `.`.

Fixes #

Checklist:
- [ ] Pull request against a `develop` branch.
- [ ] Tests added.

- 
- 
-


krzyzanowskim:
It's not an enum on master. It's not released yet though

frankdilo:
So the version I got with pod install (no version specified) is not the one on master?

PR #671, opened 2018-10-19
DON'T MERGE - Example of an improved (50%) performance for MD5

Joannis:
Not final. Would need further tweaking on the APIs and a little on the internals (specifically the class wrapper). Mainly here for demonstration purposes of how I'd improve the performance of hashes.

Original implementation of MD5 on release compilation: ~126-140 ms / run
Proposed implementation of MD5 on release compilation: 85-100ms / run

EDIT:
I must admit I didn't expect arrays/array-slices (rather than pointers) for input to be as performant as they appear now. There's no significant difference.


Joannis:
I kinda forgot about this PR, I was hoping to discuss it with you back then :)
If you feel that it's useful you can merge it and adjust it where you think it's necessary.

PR #680, opened 2018-11-20
Correct count check for initializing UInt64 from Data

marcprux:
The current UInt64 initializer will crash when the collection contains fewer than 8 bytes due to a faulty count check.


krzyzanowskim:
looks like buggy copy&paste line bug. Thanks for catching it!

PR #707, opened 2019-03-24
Fix Swift 5 warning

keith:
With Swift 5 the bytes initializer is deprecated. The replacement is
backwards compatible


krzyzanowskim:
There is Swift5 branch where compiler is set to 5. I think it's addressed on branch

keith:
This is addressed on this branch, but because this is backwards compatible if you'd like you could merge this now, without merging the breaking project file changes

krzyzanowskim:
thanks!

keith:
Thanks!

PR #738, opened 2019-07-22
Include minimum deployment targets in Swift 5 package

drewag:
CryptoSwift supports all the way back to iOS 8, macOS 10.10, and tvOS 9.

### DIFF
--- a/Package@swift-5.swift
+++ b/Package@swift-5.swift
@@ -4,6 +4,9 @@ import PackageDescription
 
 let package = Package(
     name: "CryptoSwift",
+    platforms: [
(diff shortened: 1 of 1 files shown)


Package@swift-5.swift
```diff
@@ -4,6 +4,9 @@ import PackageDescription
 
 let package = Package(
     name: "CryptoSwift",
+    platforms: [
```
krzyzanowskim: it needs 'swift-tools-version:5.1', isn't it?

drewag: Nope, platforms is supported in 5.0.

krzyzanowskim:
thank you!

PR #753, opened 2019-09-30
Changed access level to additional Array initializer 

sp3esu:
This initializer is not a part of CryptoSwift business logic so it should not be public. Moreover it's used only internally in one place and it's not mentioned in documentation.

Story behind:
I was moving some macOS code to serverside and noticed code dependency to CryptoSwift in module which didn't have anything in common with encryption (but in general CryptoSwift was used in the project). Dependency was only on this extension and it was introduced accidentally (because XCode code suggestions).

Library API should be clean and contain only methods related to value delivered by it.

### DIFF
--- a/Sources/CryptoSwift/Array+Extension.swift
+++ b/Sources/CryptoSwift/Array+Extension.swift
@@ -14,7 +14,7 @@
 //
 
 extension Array {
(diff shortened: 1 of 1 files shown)


krzyzanowskim:
fair enough. thanks!

PR #758, opened 2019-10-14
fix Swift 5.0 build

Fryie:
Checklist:
- [x] Correct file headers (see CONTRIBUTING.md). [No new files added]
- [ ] Formatted with [SwiftFormat](https://github.com/nicklockwood/SwiftFormat). [No, maybe I'm doing it wrong, but running SwiftFormat on the code modifies a lot of files I didn't even touch.]
- [ ] Tests added [not feasible for this case, one would need a CI with Linux and Swift 5.0; but I did test it in a docker image with Swift 5.0]

- The 1.1.1 release still doesn't compile on Linux with Swift 5.0 because of an implicit return which was only added to Swift 5.1. This PR fixes that issue and the code should again compile on Linux under Swift 5.0.

### DIFF
--- a/Sources/CryptoSwift/Cryptors.swift
+++ b/Sources/CryptoSwift/Cryptors.swift
@@ -37,6 +37,6 @@ extension Cryptors {
   /// Convenience helper that uses `Swift.RandomNumberGenerator`.
   /// - Parameter count: Length of array
     public static func randomIV(_ count: Int) -> Array<UInt8> {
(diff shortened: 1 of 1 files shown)


krzyzanowskim:
Thanks. This fix is already released in v1.1.1

krzyzanowskim:
ups. this is about "return" uh. Swift is so annoying

PR #763, opened 2019-10-25
To hex string

valeriyvan:
Checklist:
- [x ] Tests added.

- changes implementation of toHexString() from [UInt8] extension 10 times increasing performance


valeriyvan:
I was sure variant with format will win. But no, reduce is the winner. But another reduce.

krzyzanowskim:
I think I already had version with format ;)

PR #766, opened 2019-10-25
WIP: Gets rid of very strange looking allocations

valeriyvan:
Checklist:
- [ ] Tests added.

- Gets rid of very strange looking allocations where plain arrays could be used. Whenever possible `Array.init(unsafeUninitializedCapacity:,initializingWith:)` is used.

### DIFF
--- a/Sources/CryptoSwift/AES.swift
+++ b/Sources/CryptoSwift/AES.swift
@@ -168,12 +168,7 @@ public final class AES: BlockCipher {
     var b3 = b30 | b31 | b32 | b33
 
     let tLength = 4
-    let t = UnsafeMutablePointer<UInt32>.allocate(capacity: tLength)
-    t.initialize(repeating: 0, count: tLength)
-    defer {
-      t.deinitialize(count: tLength)
-      t.deallocate()
-    }
+    var t = Array<UInt32>(repeating: 0, count: tLength)
 
     for r in 0..<rounds - 1 {
       t[0] = b0 ^ rk[r][0]
@@ -263,12 +258,7 @@ public final class AES: BlockCipher {
     var b3 = b30 | b31 | b32 | b33
 
     let tLength = 4
-    let t = UnsafeMutablePointer<UInt32>.allocate(capacity: tLength)
-    t.initialize(repeating: 0, count: tLength)
-    defer {
-      t.deinitialize(count: tLength)
-      t.deallocate()
-    }
+    var t = Array<UInt32>(repeating: 0, count: tLength)
 
     for r in (2...rounds).reversed() {
       t[0] = b0 ^ rk[r][0]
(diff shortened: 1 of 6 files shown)


krzyzanowskim:
did you measure the performance diff? allocations are there to a ovoid costly CopyOnWrite that happens when array is mutated

valeriyvan:
> did you measure the performance diff? 

No, I didn't. But I will.

> allocations are there to a ovoid costly CopyOnWrite that happens when array is mutated

I doubt if in any of refactored places COW is involved. But definitely these cases worth a little bit more attention.

Sources/CryptoSwift/SHA1.swift
```diff
@@ -36,12 +36,7 @@ public final class SHA1: DigestType {
   fileprivate func process(block chunk: ArraySlice<UInt8>, currentHash hh: inout ContiguousArray<UInt32>) {
     // break chunk into sixteen 32-bit words M[j], 0 ≤ j ≤ 15, big-endian
     // Extend the sixteen 32-bit words into eighty 32-bit words:
-    let M = UnsafeMutablePointer<UInt32>.allocate(capacity: 80)
-    M.initialize(repeating: 0, count: 80)
-    defer {
-      M.deinitialize(count: 80)
-      M.deallocate()
-    }
+    var M = Array<UInt32>(repeating: 0, count: 80)
```
krzyzanowskim: M is mutated. I think this is an regression

Sources/CryptoSwift/AES.swift
```diff
@@ -263,12 +258,7 @@ public final class AES: BlockCipher {
     var b3 = b30 | b31 | b32 | b33
 
     let tLength = 4
-    let t = UnsafeMutablePointer<UInt32>.allocate(capacity: tLength)
-    t.initialize(repeating: 0, count: tLength)
-    defer {
-      t.deinitialize(count: tLength)
-      t.deallocate()
-    }
+    var t = Array<UInt32>(repeating: 0, count: tLength)
```
krzyzanowskim: mutated below

krzyzanowskim:
> I doubt if in any of refactored places COW is involved.

It looks like it does. Please take another look.

valeriyvan:
You are right.

AES encryption/decryption shows ~10% performance degrade:

```
             before     after refactor
AES encrypt  4.578 sec  5.199 sec
             4.595 sec  5.093 sec
             4.561 sec  5.015 sec
             --------------------
     average 4.578 sec  5.102 sec ~11% degrade

AES decrypt  5.237 sec  5.659 sec
             5.316 sec  5.739 sec
             5.118 sec  5.743 sec
             --------------------
     average 5.224 sec  5.714 sec ~9% degrade
```

There're no performance tests for SHA1, SHA2 and SHA3. I will add this missing tests to see if there a degrade as well.

Probably it makes sense to add a comment explaining a reason for that allocations.

valeriyvan:
Performance tests for SHA1, SHA2 and SHA3 were added with PR #770.
Waiting for it being merged then will continue with current PR.

krzyzanowskim:
what's the update on that?

PR #818, opened 2020-05-30
Added convenience functions to encrypt and decrypt AES with CBC mode

VatoKo:
Added simple helper functions, which are quite convenient to use to encrypt/decrypt data with AES with CBC mode


NathanFallet:
Suggestion: It could be great to add a parameter to choose the encryption algorithm (if other modes that CBC want to be used)

Something like this:
```swift
encryptAES(plainText: String, secretKey: String, mode: BlockMode, padding: Padding)
```

PR #855, opened 2021-03-01
Implementing CFB segment size

### LINKED ISSUE #500: AES encrypt, CFB mode, segment_size support?
Server side (python) required encrypt with segment_size:
`AES.new(key, AES.MODE_CFB, iv, segment_size=24).encrypt(plain)`
I tried many ways but cant find the segment_size support in this lib

NathanFallet:
Fixes #500 

Checklist:
- [x] Tests added.

- Adding a parameter to select to use cfb128 or cfb8 (cfb128 is used by default if nothing is specified)

Tested using _National Institute of Standards and Technology Special Publication 800-38A 2001 ED_ test vectors:
image

```swift
CFB(iv: iv) // Uses CFB128 as default
CFB(iv: iv, segmentSize: .cfb128) // CFB128
CFB(iv: iv, segmentSize: .cfb8) // CFB8
```

### DIFF
--- a/Sources/CryptoSwift/BlockMode/CFB.swift
+++ b/Sources/CryptoSwift/BlockMode/CFB.swift
@@ -23,21 +23,23 @@ public struct CFB: BlockMode {
   }
     
   public enum SegmentSize: Int {
-    case cfb8 = 8 // Encrypt byte per byte
-    case cfb128 = 128 // Encrypt 16 bytes per 16 bytes (default)
+    case cfb8 = 1 // Encrypt byte per byte
+    case cfb128 = 16 // Encrypt 16 bytes per 16 bytes (default)
   }
 
   public let options: BlockModeOption = [.initializationVectorRequired, .useEncryptToDecrypt]
   private let iv: Array<UInt8>
   private let segmentSize: SegmentSize
+  public let customBlockSize: Int?
 
   public init(iv: Array<UInt8>, segmentSize: SegmentSize = .cfb128) {
     self.iv = iv
     self.segmentSize = segmentSize
+    self.customBlockSize = segmentSize.rawValue
   }
 
   public func worker(blockSize: Int, cipherOperation: @escaping CipherOperationOnBlock, encryptionOperation: @escaping CipherOperationOnBlock) throws -> CipherModeWorker {
-    if self.iv.count != blockSize {
+    if self.iv.count != AES.blockSize {
       throw Error.invalidInitializationVector
     }
 
@@ -71,13 +73,12 @@ struct CFBModeWorker: BlockModeWorker {
     }
     // CFB8
     else if segmentSize == .cfb8 {
-      for i in 0 ..< plaintext.count {
-        guard let ciphertext = cipherOperation(prev ?? iv) else {
-          return Array(plaintext)
-        }
-        self.prev = Array((prev ?? iv).dropFirst()) + [plaintext[i] ^ ciphertext[0]]
+      guard let ciphertext = cipherOperation(prev ?? iv) else {
+        return Array(plaintext)
       }
-      return Array(self.prev ?? [])
+      let result = [plaintext[0] ^ ciphertext[0]]
+      self.prev = (prev ?? iv).dropFirst() + result
+      return result
     }
     return Array(plaintext) // Unsupported segment size
   }
@@ -94,15 +95,11 @@ struct CFBModeWorker: BlockModeWorker {
     }
     // CFB8
     else if segmentSize == .cfb8 {
-      var result: Array<UInt8> = []
-      for i in 0 ..< ciphertext.count {
-        guard let plaintext = cipherOperation(prev ?? iv) else {
-          return Array(ciphertext)
-        }
-        self.prev = Array((prev ?? iv).dropFirst()) + [ciphertext[i]]
-        result.append(ciphertext[i] ^ plaintext[0])
+      guard let plaintext = cipherOperation(prev ?? iv) else {
+        return Array(ciphertext)
       }
-      return result
+      self.prev = (prev ?? iv).dropFirst() + [ciphertext[0]]
+      return [ciphertext[0] ^ plaintext[0]]
     }
     return Array(ciphertext) // Unsupported segment size
   }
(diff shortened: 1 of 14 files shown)


Sources/CryptoSwift/BlockMode/CFB.swift
```diff
@@ -21,20 +21,29 @@ public struct CFB: BlockMode {
...
+    
+  public enum SegmentSize: Int {
+    case cfb8 = 1 // Encrypt byte per byte
+    case cfb128 = 16 // Encrypt 16 bytes per 16 bytes (default)
+  }
 
   public let options: BlockModeOption = [.initializationVectorRequired, .useEncryptToDecrypt]
   private let iv: Array<UInt8>
+  private let segmentSize: SegmentSize
+  public let customBlockSize: Int?
```
krzyzanowskim, line 33: the block size can be either 16 or 1, if such can you elaborate on why it should accept any block size value?

I see it as a 
```swift
blockSize = SegmentSize.cfb128.rawValue
```

Sources/CryptoSwift/BlockMode/CFB.swift
```diff
@@ -21,20 +21,29 @@ public struct CFB: BlockMode {
...
-  public init(iv: Array<UInt8>) {
+  public init(iv: Array<UInt8>, segmentSize: SegmentSize = .cfb128) {
     self.iv = iv
+    self.segmentSize = segmentSize
+    self.customBlockSize = segmentSize.rawValue
   }
 
   public func worker(blockSize: Int, cipherOperation: @escaping CipherOperationOnBlock, encryptionOperation: @escaping CipherOperationOnBlock) throws -> CipherModeWorker {
-    if self.iv.count != blockSize {
+    if !(self.iv.count == blockSize || (segmentSize == .cfb8 && self.iv.count == AES.blockSize)) {
```
krzyzanowskim, line 42: `BlockMode` should stay cipher-independent whenever possible.  Can we get rid of `AES.blockSize` here?

Sources/CryptoSwift/BlockMode/CFB.swift
```diff
@@ -43,28 +52,55 @@ struct CFBModeWorker: BlockModeWorker {
...
+      self.prev = xor(plaintext, ciphertext.slice)
+      return Array(self.prev ?? [])
+    }
+    // CFB8
+    else if segmentSize == .cfb8 {
+      guard let ciphertext = cipherOperation(prev ?? iv) else {
+        return Array(plaintext)
+      }
+      let result = [Array(plaintext)[0] ^ Array(ciphertext)[0]]
+      self.prev = Array((prev ?? iv).dropFirst()) + [result[0]]
```
krzyzanowskim, line 80: isn't it... slow? Did you check maybe how it compares to CFB128?

NathanFallet: I understand it is slower than CFB128, but CFB8 is designed to encrypt byte per byte, and so the IV is shifted with each new encrypted byte each time. That's also why the block size is 1 byte and not 16 bytes.

Sources/CryptoSwift/BlockMode/CFB.swift
NathanFallet: Here `customBlockSize` has a getter purpose for `BlockMode`, and is set to `segmentSize.rawValue` in `init` method. (In case it is not nil, it overrides default AES blockSize of 16 bytes)

Sources/CryptoSwift/BlockMode/CFB.swift
NathanFallet: The problem here is that `blockSize` is 1 when `segmentSize` is `.cfb8`, and the IV should stay 16 bytes long (for AES), so I can't compare `iv.count` to `blockSize`. Maybe we could add a blockCipherSize arg to know if it's AES or something else? (or keep it like that)

krzyzanowskim: hm... why do we need `BlockMode.customBlockSize`? the block size is a parameter already. I don't see much value in this property - I don't see a relation to segment size hm...

how about remove `customBlockMode`, instead store block mode in `CFB` and use the stored value here?

NathanFallet: I need this constant in the `AES.swift` file for example, because CFB8 is not working on 16 bytes like default AES but 1 byte instead (else there are errors when the entry size is not a multiple of 16 bytes)
https://github.com/krzyzanowskim/CryptoSwift/blob/085709b6a4e33b60fa0b4a4629655fdd1ba5644e/Sources/CryptoSwift/AES.swift#L499

krzyzanowskim: this value goes down to block mode anyway, so it can be checked here and leave the block mode itself unchanged. Another thing, in CFB, let's check if passed `blockSize` is 1 byte instead as it's the only compatible.

NathanFallet: If I remove the `customBlockSize` constant, test fails:
```
Test Case '-[CryptoSwiftTests.AESTests testAESEncryptCFB8]' started.
/Users/nathanfallet/git/CryptoSwift/Tests/CryptoSwiftTests/AESTests.swift:174: error: -[CryptoSwiftTests.AESTests testAESEncryptCFB8] : XCTAssertEqual failed: ("[59, 121, 66]") is not equal to ("[59, 121, 66, 76, 156, 13, 212, 54, 186, 206, 158, 14, 212, 88, 106, 79, 50, 185]") - encryption failed
/Users/nathanfallet/git/CryptoSwift/Tests/CryptoSwiftTests/AESTests.swift:176: error: -[CryptoSwiftTests.AESTests testAESEncryptCFB8] : XCTAssertEqual failed: ("[107]") is not equal to ("[107, 193, 190, 226, 46, 64, 159, 150, 233, 61, 126, 17, 115, 147, 23, 42, 174, 45]") - decryption failed
```
So this constant is necessary for CFB8 to work, here in `AES`'s `Cryptors` extension:

https://github.com/krzyzanowskim/CryptoSwift/pull/855/files#diff-68bfbf24a9e4229282e0d21efc2b8afb5d4d3db5ad131cc6a185d613f0da78e8R20
https://github.com/krzyzanowskim/CryptoSwift/pull/855/files#diff-68bfbf24a9e4229282e0d21efc2b8afb5d4d3db5ad131cc6a185d613f0da78e8R29

krzyzanowskim: ok, let's get it in, and I want to play with it a bit, maybe I'll figure how to make it out of it. Great work btw.

krzyzanowskim (approved):
Thank you! it's awesome

PR #862, opened 2021-04-23
Support watchos in spm

ojii:
The [podspec](https://github.com/krzyzanowskim/CryptoSwift/blob/1adeed4bc54f2b3e5feba2af2a55a3cac5c9737f/CryptoSwift.podspec#L15) supports watchos, but the spm metadata doesn't.


krzyzanowskim:
thank you!

PR #872, opened 2021-05-25
[SHA2.swift] warning: default will never be executed

aiden-leong:
remove exhausted default case


Sources/CryptoSwift/SHA2.swift
```diff
@@ -316,8 +316,6 @@ extension SHA2: Updatable {
             self.process32(block: chunk, currentHash: &self.accumulatedHash32)
           case .sha384, .sha512:
             self.process64(block: chunk, currentHash: &self.accumulatedHash64)
-          @unknown default:
-            preconditionFailure()
```
krzyzanowskim, line 320: Screenshot 2021-06-18 at 11 58 00

It doesn't look right. It is needed to silence the warning

krzyzanowskim (changes requested):
Version 12.5 report warning without default cases.

PR #873, opened 2021-05-26
Optionality is removed from return type

evnik:
It seems optionality is not needed after 47eefd0


krzyzanowskim (approved):
great spot, thank you!

PR #886, opened 2021-07-17
Adds support for SHA3 when using HMAC

R4v3nPr0:
This pull request adds support for SHA3 on HMAC

### DIFF
--- a/Sources/CryptoSwift/HMAC.swift
+++ b/Sources/CryptoSwift/HMAC.swift
@@ -20,7 +20,7 @@ public final class HMAC: Authenticator {
   }
 
   public enum Variant {
-    case sha1, sha256, sha384, sha512, md5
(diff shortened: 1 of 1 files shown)


Sources/CryptoSwift/HMAC.swift
```diff
@@ -20,7 +20,7 @@ public final class HMAC: Authenticator {
   }
 
   public enum Variant {
-    case sha1, sha256, sha384, sha512, md5
+    case sha1, sha256, sha384, sha512, sha3_224, sha3_256, sha3_384, sha3_512, md5
```
krzyzanowskim, line 23: let's simplify the API a bit by using a parameter here, I'm thinking about reuse SHA3.Variant here like this:

```swift
case sha3(SHA3.Variant)
```

krzyzanowskim (changes requested):
Thank you! it's awesome. Would you mind adjust API slightly, as I suggested in comment?

PR #887, opened 2021-08-17
RSA

### LINKED ISSUE #63: RSA?
I would really love to see the addition of RSA public key encryption methods in this framework, given that there aren't many Swift implementations out there right now.

NathanFallet:
Fixes #63

Checklist:
- [x] Tests added.

- [x] Using [BigInt](https://github.com/attaswift/BigInt) for RSA computation
- [x] Encrypt/Decrypt content
- [x] Generate key pair
- [x] Add documentation in README
- ~~Load/Export from/to PEM~~ (see #892)


NathanFallet:
Asked on StackOverflow for help with number computation: https://stackoverflow.com/questions/68883661/big-number-computation-in-swift-for-rsa-implementation

CryptoSwift.xcodeproj/project.pbxproj
```diff
@@ -1192,6 +1208,25 @@
 			defaultConfigurationName = Release;
 		};
 /* End XCConfigurationList section */
+
```
tgymnich, line 1211: maybe don't add BigInt as a dependency here, but use cocoa pods instead with the post_install hook.

krzyzanowskim (approved):

PR #915, opened 2022-04-15
Fix Xcode 13.3.1 Cycle error

mime29:
- When using CryptoSwift as a dependency for frameworks with Xcode 13.3.1 we can observe the following error:

```
error: Cycle inside CryptoSwift; building could produce unreliable results. This usually can be resolved by moving the target's Headers build phase before Compile Sources.
```

This PR fixes this compilation issue by updating the Build Phases and setting the Headers compilation before the Sources compilation.

This problem seems to be linked to the following bug report: https://developer.apple.com/forums/thread/700997


krzyzanowskim:
thank you!

mime29:
You're welcome ! 
Could you publish a new release version please?

PR #921, opened 2022-07-01
RSA + primes

btoms20:
Fixes #
The `init(keySize:Int)` method now throws, introducing a breaking change from the previous implementation. The previous initializer assumed the BigInt inverse operation would never result in nil (and it shouldn't) but if it ever did, you'd get back a public key instead of a private key (which is sort of confusing).

Checklist:
- [x] Tests updated.

- Added an optional private `primes` constant that holds on to the primes (p & q) that are used to generate the RSA key's private exponent. This is an important first step to being able to support DER and PEM importing and exporting as they both require the primes to be bundled within the Private key's DER format.

### DIFF
--- a/Sources/CryptoSwift/RSA.swift
+++ b/Sources/CryptoSwift/RSA.swift
@@ -25,6 +25,8 @@ public final class RSA {
   public enum Error: Swift.Error {
     /// No private key specified
     case noPrivateKey
+    /// Failed to calculate the inverse e and phi
+    case invalidInverseNotCoprimes
   }
   
   /// RSA Modulus
@@ -39,6 +41,9 @@ public final class RSA {
   /// The size of the modulus, in bits
   public let keySize: Int
   
+  /// The underlying primes used to generate the Private Exponent
+  private let primes:(p:BigUInteger, q:BigUInteger)?
+
   /// Initialize with RSA parameters
   /// - Parameters:
   ///   - n: The RSA Modulus
@@ -48,7 +53,8 @@ public final class RSA {
     self.n = n
     self.e = e
     self.d = d
-    
+    self.primes = nil
+
     self.keySize = n.bitWidth
   }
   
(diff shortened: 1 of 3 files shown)


Sources/CryptoSwift/RSA.swift
```diff
@@ -21,24 +21,29 @@ import Foundation
...
   /// RSA Private Exponent
   public let d: BigUInteger?
-  
+
   /// The size of the modulus, in bits
   public let keySize: Int
-  
+
+  /// The underlying primes used to generate the Private Exponent
+  private let primes: (p: BigUInteger, q: BigUInteger)?
```
NathanFallet, line 45: Why are you storing p and q? They are unused.

NathanFallet (commented):
Good except you need to explain why you store p and q

btoms20:
### To clear things up and add a bit of context
This PR and PR #922 are both working towards Issue #958 Instead of submitting one giant PR with a ton of changes, I figured I'd split them up into more manageable parts...
1) #921  - Update RSA Initializers to include one that accepts primes for DER/PEM imports, and to hold onto the generated primes for DER/PEM exports
2) #922 - Which introduces the DERCodable Protocol that cryptographic keys can conform to in order to add import/export behavior via the DERDecodable and DEREncodable protocols. 
3) Submit a future PR with RSA conforming to DERCodable (code snippet included below)
4) Submit a future PR introducing PEM imports and exports

### As to why we store the Primes
The values `p` and `q` are used when exporting and/or importing the RSA key as a DER and/or PEM file type. I presented this change alone due to it including a change to the `RSA` initializers. I thought smaller incremental changes would be more manageable than one huge PR.

These values would eventually be used in order for RSA to conform to the DERCodable Protocol presented in PR #922 

The following is an example of exporting the Private RSA Key as an ASN1 DER encoded external representation following the [IETF RFC2313 Spec](https://datatracker.ietf.org/doc/html/rfc2313#section-7.2).

```swift
extension RSA: DEREncodable {
  /// The DER representation of this private key
  ///
  /// [IETF Spec RFC2313](https://datatracker.ietf.org/doc/html/rfc2313#section-7.2)
  /// ```
  /// ==========================
  ///  RSA PrivateKey Structure
  /// ==========================
  ///
  /// RSAPrivateKey ::= SEQUENCE {
  ///   version           Version,
  ///   modulus           INTEGER,  -- n
  ///   publicExponent    INTEGER,  -- e
  ///   privateExponent   INTEGER,  -- d
  ///   prime1            INTEGER,  -- p
  ///   prime2            INTEGER,  -- q
  ///   exponent1         INTEGER,  -- d mod (p-1)
  ///   exponent2         INTEGER,  -- d mod (q-1)
  ///   coefficient       INTEGER,  -- (inverse of q) mod p
  ///   otherPrimeInfos   OtherPrimeInfos OPTIONAL
  /// }
  /// ```
  func privateKeyDER() throws -> Array<UInt8> {
    // Make sure we have a private key
    guard let d = d else { throw RSA.Error.noPrivateKey }
    // Make sure we have access to our primes
    guard let primes = primes else { throw RSA.Error.noPrimes }
    // Make sure we can calculate our coefficient (inverse of q mod p)
    guard let coefficient = primes.q.inverse(primes.p) else { throw RSA.Error.unableToCalculateCoefficient }
  
    let bitWidth = self.keySize / 8
    let paramWidth = bitWidth / 2
    // Structure the data
    let mod = self.n.serialize()
    let privateKeyAsnNode: ASN1.Node =
      .sequence(nodes: [
        .integer(data: Data(hex: "0x00")),
        .integer(data: DER.i2ospData(x: mod.bytes, size: bitWidth)),
        .integer(data: DER.i2ospData(x: self.e.serialize().bytes, size: 3)),
        .integer(data: DER.i2ospData(x: d.serialize().bytes, size: bitWidth)),
        .integer(data: DER.i2ospData(x: primes.p.serialize().bytes, size: paramWidth)),
        .integer(data: DER.i2ospData(x: primes.q.serialize().bytes, size: paramWidth)),
        .integer(data: DER.i2ospData(x: (d % (primes.p - 1)).serialize().bytes, size: paramWidth)),
        .integer(data: DER.i2ospData(x: (d % (primes.q - 1)).serialize().bytes, size: paramWidth)),
        .integer(data: DER.i2ospData(x: coefficient.serialize().bytes, size: paramWidth))
      ])
  
    // Encode and return the data
    return ASN1.Encoder.encode(privateKeyAsnNode)
  }
}
```
 
The above code snippet requires the primes and the ASN1 Encoder from PR #922 in order to export the RSA Private Key.

If you would prefer that I close both this PR and PR #922 and instead submit all of the DER / PEM & RSA conformance code in one go, I can do that as well. Just let me know. 

Thanks

NathanFallet (approved):

NathanFallet:
@btoms20 Perfect!

Issue #892, opened 2021-09-26
Add PEM import/export to RSA encryption


NathanFallet:
I made a first PR to implement RSA (see #63 and #887), but did not have time to implement PEM related features.
So I plan to make another PR to fix this issue separately.

PR #922, opened 2022-07-09
Simple ASN1 Encoder and Decoder & DERCodable Protocol

btoms20:
Working towards #892

Checklist:
- [x] Tests added. (ASN1 specific tests added)

- This PR implements a simple ASN1 Encoder and Decoder. Abstract Syntax Notation (ASN) is often used along side Distinguished Encoding Rules (DER) for [structuring cryptographic data](https://datatracker.ietf.org/doc/html/rfc2313#section-7) before, and after, processing it (processing includes importing, exporting, signatures & verification, encryption & decryption, etc). 
- This PR also introduces a `DERDecodable`, `DEREncodable` and `DERCodable` Protocol set that cryptographic keys, such as `RSA`, can conform to in order to support Importing and Exporting keys that are structured as an ASN1 DER representation.
- Apple's `Security` framework exports ASN1 DER representations when using the `SecKeyCopyExternalRepresentation` function. This PR would allow for exchanging / sharing RSA keys between CryptoSwift and Apple's `Security` framework. The `DEREncodable` Protocol exposes  two functions `externalRepresentation()` and `publicKeyExternalRepresentation()` that should result in the same external representation as Apple's `Security` `SecKeyCopyExternalRepresentation` function.
- This PR sets the groundwork for supporting the importing and exporting of PEM files.
- This PR would also enable `RSA` Signature and Verification functionality.

### DIFF
--- a/Sources/CryptoSwift/ASN1/ASN1.swift
+++ b/Sources/CryptoSwift/ASN1/ASN1.swift
@@ -1,0 +1,95 @@
+//
+//  CryptoSwift
+//
+//  Copyright (C) Marcin Krzyżanowski <marcin@krzyzanowskim.com>
+//  This software is provided 'as-is', without any express or implied warranty.
+//
+//  In no event will the authors be held liable for any damages arising from the use of this software.
+//
+//  Permission is granted to anyone to use this software for any purpose,including commercial applications, and to alter it and redistribute it freely, subject to the following restrictions:
+//
+//  - The origin of this software must not be misrepresented; you must not claim that you wrote the original software. If you use this software in a product, an acknowledgment in the product documentation is required.
+//  - Altered source versions must be plainly marked as such, and must not be misrepresented as being the original software.
+//  - This notice may not be removed or altered from any source or binary distribution.
+//
+//  ASN1 Code inspired by Asn1Parser.swift from SwiftyRSA
+
+import Foundation
+
+enum ASN1 {
+  internal enum IDENTIFIERS: UInt8, Equatable {
+    case SEQUENCE = 0x30
+    case INTERGER = 0x02
+    case OBJECTID = 0x06
+    case NULL = 0x05
+    case BITSTRING = 0x03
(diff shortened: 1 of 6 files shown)


btoms20:
Hey @krzyzanowskim! First off, awesome work with the CryptoSwift framework! 

I have the code for `RSA:DERCodable` conformance, and RSA Signatures and Signature Verification ready to submit once this PR is approved / merged.

I also have a preliminary implementation of RSA PEM support (including encrypted PEM, imports & exports) which would help close out issue  #892 and provide additional Key interoperability between frameworks (like Openssl, Apple's `Security`, etc...). Encrypted PEMs would also provide a secure method of saving keys between CryptoSwift sessions. 👍 

Let me know if anything in this PR needs to be changed in order to be approved.

Thanks!

krzyzanowskim:
thank you! I'm on vacation, but will back to it after next week.

NathanFallet (changes requested):
You need to add tests.

btoms20:
Okay, will do! 👍

btoms20:
Hey @NathanFallet, I added some ASN1 tests located at c242291adc75a44cfc06cafd290a9532c8df4246. 

I also added some additional documentation and notes regarding the current status of the ASN.1 implementation. It's an internal tool (not to be used by the end user) but only from within CryptoSwift for specific, tested, use cases, such as DER / PEM encoding and decoding.  

Let me know if you have any additional questions, concerns, changes, etc...

 Thanks

NathanFallet:
If it is implemented in the library for internal use, why can't it be used publicly? Is it too hard to maintain?
I will take a look to your changes

btoms20:
ASN1 is a giant standard, consisting of many sub standards and the implementation proposed in this PR is a small subset of the main standard useful for handling Cryptographic keys only. I think exposing a partial ASN1 implementation to the public facing API would be confusing and potentially unsafe.

PR #978, opened 2022-09-01
RSA Raw Signature Support and Key Size Byte Calculation Fix

btoms20:
Fixes:
When calculating the key's size in bytes, we weren't taking into consideration any potential remaining bits. This resulted in keys who's size in bits weren't multiples of 8 failing to handle signatures correctly (the signing and verification methods would hang indefinitely).

Checklist:
- [x] Tests added.
   - Added a 1023 bit RSA test fixture

- Added support for the `.raw` and `.digest_pkcs1v15_RAW` signature variants
   - These variants are semantically and functionally similar to the ones in Apple's `Security` framework.
- Added a `keySizeBytes:Int` parameter to the `RSA` class.
   - We were doing a lot of repetitive bit to byte calculations and not properly accounting for bit remainders.
   - We now calculate the minimum number of bytes needed to represent the keys modulus once upon initialization.

### DIFF
--- a/Sources/CryptoSwift/RSA/RSA+Cipher.swift
+++ b/Sources/CryptoSwift/RSA/RSA+Cipher.swift
@@ -106,20 +106,7 @@ extension RSA {
           return bytes
         case .raw, .pksc1v15:
           // Format the encrypted bytes before returning
-          //var bytes = bytes
-//          if bytes.isEmpty {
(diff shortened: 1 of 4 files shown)


NathanFallet:
I'm going to take a look as soon as possible. Thanks again for your contribution!

NathanFallet (approved):
Looks good to me.
Also great to fix this issue with 1023 or 2047 bits length.

PR #989, opened 2022-10-11
Fix "Default will never be executed" warnings

jpsim:
when building without library evolution enabled.

Similar to 9850f4cb23152322dbb85c897c7b83453f191b18.

Checklist:
- [ ] Tests added. N/A

- Remove `@unknown default` switch conditions because these cause compiler warnings when library evolution is not enabled.


jpsim:
As pointed out by @keith, this shouldn't ever be needed even with library evolution enabled because the enum is defined in the same module as this switch statement.

krzyzanowskim:
Xcode 14.0.1

Screen Shot 2022-10-11 at 19 53 48@2x

I'm confused now

jpsim:
Try `xed Package.swift` and cmd-B.

krzyzanowskim:
so... it needs `@frozen` for `RSA.RSAEncryptionVariant`. Do you mind adding it in this PR?

krzyzanowskim:
thank you!

jpsim:
Thanks for the quick review, now please go back to spending time with your kids!

PR #1013, opened 2023-04-21
Fix warnings with Swift 5.8

jpsim:
There are a few warnings when compiling `main` with Swift 5.8:

```
Sources/CryptoSwift/Scrypt.swift:124:7 'assign(from:count:)' is deprecated: renamed to 'update(from:count:)'
Sources/CryptoSwift/String+Extension.swift:21:5 Instance method 'data(using:allowLossyConversion:)' cannot be used in an '@inlinable' function because 'Foundation' was not imported by this file; this is an error in Swift 6
Sources/CryptoSwift/String+Extension.swift:21:24 Struct 'Encoding' cannot be used in an '@inlinable' function because 'Foundation' was not imported by this file; this is an error in Swift 6
Sources/CryptoSwift/String+Extension.swift:21:33 Static property 'utf8' cannot be used in an '@inlinable' function because 'Foundation' was not imported by this file; this is an error in Swift 6
```

This fixes them all while continuing to support previous Swift compiler versions.

Fixes #

Checklist:
- [ ] Tests added.

- Fix warnings with Swift 5.8

### DIFF
--- a/Sources/CryptoSwift/Scrypt.swift
+++ b/Sources/CryptoSwift/Scrypt.swift
@@ -121,7 +121,11 @@ private extension Scrypt {
 
     /* 1: X <-- B */
     let typedBlock = block.assumingMemoryBound(to: UInt32.self)
+#if compiler(>=5.8)
+    X.update(from: typedBlock, count: 32 * self.r)
+#else
     X.assign(from: typedBlock, count: 32 * self.r)
+#endif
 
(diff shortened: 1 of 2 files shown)


krzyzanowskim (approved):
LGTM thank you!

jpsim:
Thanks for the quick merge!

PR #1014, opened 2023-05-03
Modify the access level of `RSA(n:e:d:p:q:)` initializer to public

zssz:
Fixes #

Checklist:
- [x] Tests added.

- Modify the access level of `RSA(n:e:d:p:q:)` initializer to public, so that users can set the prime components p and q. This change is necessary when working with RSA private keys that need to be initialized from scratch, such as from a [JWK](https://www.rfc-editor.org/rfc/rfc7517) representation.

### DIFF
--- a/Sources/CryptoSwift/RSA/RSA.swift
+++ b/Sources/CryptoSwift/RSA/RSA.swift
@@ -119,7 +119,7 @@ public final class RSA: DERCodable {
   ///   - d: The RSA Private Exponent
   ///   - p: The 1st Prime used to generate the Private Exponent
   ///   - q: The 2nd Prime used to generate the Private Exponent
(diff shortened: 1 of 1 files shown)


NathanFallet:
In case you make this initializer public, I think you should move some of the check from `init(privateDER der: Array<UInt8>)` to this initializer, to be sure that primes are valid.

NathanFallet (changes requested):
In case you make this initializer public, I think you should move some of the checks from `init(privateDER der: Array<UInt8>)` to this initializer, to be sure that primes are valid.

zssz:
Thank you, @NathanFallet, for the review. I have made the changes you suggested and committed them.

NathanFallet (approved):

PR #1021, opened 2023-09-11
PKCS7Padding: Remove unreachable if branch

matthiasgeihs:
**Changes proposed in this pull request:**
`padding` could never be zero as `blockSize > 0` and `bytes.count % blockSize < blockSize`.
Hence, we remove the unreachable if branch.
Addresses https://github.com/krzyzanowskim/CryptoSwift/discussions/1020.

**Checklist:**
- [ ] Tests added.

**Note:**
- No tests were added because change only removes unreachable code.
- `swiftformat` was run on the file that contains changes. However, running `swiftformat` on the entire repository would create a lot of additional changes, as many files don't seem to be formatted.

### DIFF
--- a/Sources/CryptoSwift/PKCS/PKCS7Padding.swift
+++ b/Sources/CryptoSwift/PKCS/PKCS7Padding.swift
@@ -29,13 +29,8 @@ struct PKCS7Padding: PaddingProtocol {
   func add(to bytes: Array<UInt8>, blockSize: Int) -> Array<UInt8> {
     let padding = UInt8(blockSize - (bytes.count % blockSize))
     var withPadding = bytes
-    if padding == 0 {
-      // If the original data is a multiple of N bytes, then an extra block of bytes with value N is added.
-      withPadding += Array<UInt8>(repeating: UInt8(blockSize), count: Int(blockSize))
-    } else {
-      // The value of each added byte is the number of bytes that are added
-      withPadding += Array<UInt8>(repeating: padding, count: Int(padding))
-    }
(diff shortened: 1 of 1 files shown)


Sources/CryptoSwift/PKCS/PKCS7Padding.swift
```diff
@@ -29,13 +29,8 @@ struct PKCS7Padding: PaddingProtocol {
...
     var withPadding = bytes
-    if padding == 0 {
-      // If the original data is a multiple of N bytes, then an extra block of bytes with value N is added.
-      withPadding += Array<UInt8>(repeating: UInt8(blockSize), count: Int(blockSize))
-    } else {
-      // The value of each added byte is the number of bytes that are added
-      withPadding += Array<UInt8>(repeating: padding, count: Int(padding))
-    }
+    // The value of each added byte is the number of bytes that are added
+    withPadding += Array<UInt8>(repeating: padding, count: Int(padding))
```
NathanFallet, line 33: You're right, it cannot be zero since `a % b` is always `< b`.
But in that case you can event write it as `return bytes + Array...` to remove two extra lines (and maybe use less memory, but I'm not sure it changes anything on that)

NathanFallet (changes requested):

Sources/CryptoSwift/PKCS/PKCS7Padding.swift
matthiasgeihs: applied your suggestion

PR #1023, opened 2023-09-24
Update BlockDecryptor.swift

AsifSyeed:
fatalError("Not supported") causes crash issue for this case. It would be better if we just print something or handle any other way

Fixes #

Checklist:
- [ ] Tests added.

-

### DIFF
--- a/Sources/CryptoSwift/BlockDecryptor.swift
+++ b/Sources/CryptoSwift/BlockDecryptor.swift
@@ -87,7 +87,7 @@ public class BlockDecryptor: Cryptor, Updatable {
 
   public func seek(to position: Int) throws {
     guard var worker = self.worker as? SeekableModeWorker else {
-      fatalError("Not supported")
+      print("Not supported")
     }
 
     try worker.seek(to: position)


Sources/CryptoSwift/BlockDecryptor.swift
```diff
@@ -87,7 +87,7 @@ public class BlockDecryptor: Cryptor, Updatable {
 
   public func seek(to position: Int) throws {
     guard var worker = self.worker as? SeekableModeWorker else {
-      fatalError("Not supported")
+      print("Not supported")
```
NathanFallet, line 90: The correct other way to handle it is to throw a custom exception so that you can handle the exception as you want. But for people who cares about this error, removing it with a print won't help them, because this error has a meaning that cannot be ignored for everyone.

NathanFallet (changes requested):

Sources/CryptoSwift/BlockDecryptor.swift
krzyzanowskim: totally. bo fatal error. It can be used ad a weakness to decrypt the original message (oracle attack kindof)

krzyzanowskim:
thanks you. I've made these changes https://github.com/krzyzanowskim/CryptoSwift/commit/aa90fb447c5abd148e35a019662092fd76a1df82 with throwing error

PR #1029, opened 2023-11-15
Update RSA's publicKeyDER method to support exponent any byte length

WFrost3:
Fixes #1028

> The `publicKeyDER` method for in `RSA.swift` only supports the conversion of RSA keys that have a exponent that is exactly 3 bytes long.
> 
> Most commonly, exponents are either 65537 or 3 and while an exponent of 3 [can be vulnerable](https://en.wikipedia.org/wiki/RSA_(cryptosystem)#:~:text=When%20encrypting%20with%20low%20encryption%20exponents%20(e.g.%2C%20e%20%3D%203)%20and%20small%20values%20of%20the%20m%20(i.e.%2C%20m%20%3C%20n1/e)%2C%20the%20result%20of%20me%20is%20strictly%20less%20than%20the%20modulus%20n.%20In%20this%20case%2C%20ciphertexts%20can%20be%20decrypted%20easily%20by%20taking%20the%20eth%20root%20of%20the%20ciphertext%20over%20the%20integers.) it is a valid exponent; the common exponents have byte lengths of 3 and 1. In addition, exponents can be non standard and have a byte length that is not 3 or 1.

Checklist:
- [x] Tests added.

- `publicKeyDER` updated to work for exponent of any size
- Test added for 1 byte long exponent


nathanfallet (approved):
Thank you!

PR #1047, opened 2024-04-28
Add support for fernet

juri:
Add support for [fernet]. This is based on a [library][juri-fernet] I created. If you think this would make sense inside CryptoSwift, I'd be happy to contribute this.

[fernet]: https://github.com/fernet/spec
[juri-fernet]: https://github.com/juri/fernet

Checklist:
- [x] Tests added.

### DIFF
--- a/Sources/CryptoSwift/Fernet.swift
+++ b/Sources/CryptoSwift/Fernet.swift
@@ -1,0 +1,190 @@
+import Foundation
+
+/// Fernet provides support for the [fernet](https://github.com/fernet/spec) encryption format.
+public struct Fernet {
+  let makeDate: () -> Date
+  let makeIV: (Int) -> [UInt8]
+  let signingKey: Data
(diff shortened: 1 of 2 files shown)


Sources/CryptoSwift/Fernet.swift
```diff
@@ -0,0 +1,203 @@
...
+  where C1: Collection,
+  C2: Collection,
+  C1.Element == UInt8,
+  C2.Element == UInt8 {
+  guard lhs.count == rhs.count else { return false }
+  return zip(lhs, rhs).reduce(into: 0) { output, pair in output |= pair.0 ^ pair.1 } == 0
+}
+
+private extension Data {
+  init?(base64URLData base64: Data) {
```
krzyzanowskim, line 169: CryptoSwift already has base64 utilities that can be used instead, can't it?

PR #1058, opened 2024-12-05
Support for Static Linux SDK Based on Musl Instead of Glibc

AnneBlair:
Hello,

I am currently working on a project that requires the use of a static Linux SDK. We are opting to use Musl instead of Glibc due to its excellent support for static linking and its permissive licensing, which simplifies the distribution of statically linked executables.

To ensure compatibility with Musl, I suggest a modification in the conditional compilation statements for importing C libraries. Here is the current code:

#if canImport(Darwin)
import Darwin
#elseif canImport(Glibc)
import Glibc
#elseif canImport(WinSDK)
import WinSDK
#endif

And here is the proposed change:

#if canImport(Darwin)
import Darwin
#elseif canImport(Glibc)
import Glibc
#elseif canImport(Musl)
import Musl
#elseif canImport(WinSDK)
import WinSDK
#endif

This change will allow the framework to dynamically check for the availability of Glibc or Musl, thus supporting environments built with Musl. This would greatly enhance the flexibility and usability of SwifterSwift in various Linux environments.

For further reference on static linking in Linux, you can visit the official Swift documentation: 
[Static Linux Getting Started](https://www.swift.org/documentation/articles/static-linux-getting-started.html).

Thank you for considering this enhancement.

### DIFF
--- a/Sources/CryptoSwift/BlockMode/CCM.swift
+++ b/Sources/CryptoSwift/BlockMode/CCM.swift
@@ -21,6 +21,8 @@
 import Darwin
 #elseif canImport(Glibc)
 import Glibc
+#elseif canImport(Musl)
+import Musl
 #elseif canImport(ucrt)
 import ucrt
 #elseif canImport(WASILibc)
--- a/Sources/CryptoSwift/Cryptors.swift
+++ b/Sources/CryptoSwift/Cryptors.swift
@@ -17,6 +17,8 @@
 import Darwin
 #elseif canImport(Glibc)
 import Glibc
+#elseif canImport(Musl)
+import Musl
 #elseif canImport(ucrt)
 import ucrt
 #endif
--- a/Sources/CryptoSwift/HKDF.swift
+++ b/Sources/CryptoSwift/HKDF.swift
@@ -20,6 +20,8 @@
 import Darwin
 #elseif canImport(Glibc)
 import Glibc
+#elseif canImport(Musl)
+import Musl
 #elseif canImport(ucrt)
 import ucrt
 #elseif canImport(WASILibc)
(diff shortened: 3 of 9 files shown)


nathanfallet (approved):

nathanfallet:
Any way to add tests so we are sure everything works with Musl too? (Maybe adding a workflow or something to cover this case)

AnneBlair:
Add Static Linux SDK Build Verification for Swift 6.0.2

While I have verified the build locally, I am not very familiar with the GitHub submission verification process. This pull request introduces a build verification process using the Static Linux SDK for Swift 6.0.2 on macOS environments. The key component used is the Static Linux SDK, which can be obtained from [Swift.org](https://www.swift.org/install/macos/). This ensures that our codebase remains compatible with Linux platforms using Musl as the standard library.

**Setup Steps:**

1. Install the Static Linux SDK for Swift as described in the [Swift Static Linux Getting Started Guide](https://www.swift.org/documentation/articles/static-linux-getting-started.html).
2. List the available Swift SDKs using the command:
   ```bash
   swift SDK list
   ```
3. Set the environment to use Swift 6.0.2:
   ```bash
   export PATH="/Library/Developer/Toolchains/swift-6.0.2-RELEASE.xctoolchain/usr/bin:$PATH"
   ```
4. Verify the Swift version:
   ```bash
   swift --version
   ```
   Expected output:
   ```
   Apple Swift version 6.0.2 (swift-6.0.2-RELEASE)
   Target: x86_64-apple-macosx14.0
   ```

5. Build the project using:
   ```bash
   swift build -v
   ```

This setup ensures that our Swift projects are correctly built using the specified Static Linux SDK, which is crucial for maintaining cross-platform compatibility.

krzyzanowskim (approved):

krzyzanowskim:
thank you!

PR #1060, opened 2024-12-06
fix: externalRepresentation condition to validate if key is private should be d not prime

beatt83:
Fixes #

I got to this issue recently where the library was giving me a public key DER when the key had a d and no primes, this is misleading since an RSA key can have d and no primes. This function should acess if the key is private through the d, if it requires the primes it should return an error.

As part of this PR I also made the primes public, they are already a let, making them public makes the API more accessible to deal for example if I want to build a JWK with primes included.
It is quite difficult right now if you want to expose this keys to other coding types since the primes are private, and as far as I know there is no reason for that.

Checklist:
- [ ] Tests added.

- RSA.externalRepresentation() private of public condition should be made validate through the variable d not primes.
- Change primes to public.

### DIFF
--- a/Sources/CryptoSwift/RSA/RSA.swift
+++ b/Sources/CryptoSwift/RSA/RSA.swift
@@ -61,7 +61,7 @@ public final class RSA: DERCodable {
   public let keySizeBytes: Int
 
   /// The underlying primes used to generate the Private Exponent
-  private let primes: (p: BigUInteger, q: BigUInteger)?
+  public let primes: (p: BigUInteger, q: BigUInteger)?
 
   /// Initialize with RSA parameters
   /// - Parameters:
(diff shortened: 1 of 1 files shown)


krzyzanowskim:
@beatt83 why this PR got closed?

krzyzanowskim:
I reopen and let @nathanfallet to weight in about the change.

goncalo-frade-iohk:
@krzyzanowskim don't really know why it closed. I just merged a PR on other repo, that referenced this PR.

Weird GitHub interaction.
Thx for reopening.

beatt83:
Hey @krzyzanowskim. It was a weird behavior with GitHub, since I tagged this PR in a commit and then merged on my library that PR it closed this one.

Thanks for reopening. :)

beatt83:
Hey @krzyzanowskim . Is there any update around this? :)

nathanfallet:
Sorry, I was quite busy at work. Just reviewed and merged!

PR #1061, opened 2024-12-09
Use `.sha2(.sha256)` for PBKDF2 in readme

sjudd:
.sha256 is deprecated if I'm reading this right.

Fixes #

Checklist:
- [ ] Tests added.

- Replace a deprecated call in an example in the readme with a non-deprecated equivalent.


krzyzanowskim (approved):
thank you

PR #1067, opened 2025-01-19
fix funtcion XCTestManifests.swift

futreall:


krzyzanowskim:
in fact, I don't think the XCTestManifests are needed anymore at all and this machinery is handled completely by SPM

PR #1069, opened 2025-01-28
Update LICENSE

maximevtush:
Updated the copyright year to 2025 in the LICENSE file.


nathanfallet:
Actually the number was chosen far to avoid updating it all the time. So we'll keep it as it is.

PR #1076, opened 2025-07-09
Rename the `bytes` computed property in the extension on `Data` to `byteArray`

mluisbrown:
Rename the `bytes` computed property in the extension on Data to `byteArray` to avoid clashing with new `bytes: RawSpan` property added in [SE-0456](https://github.com/swiftlang/swift-evolution/blob/main/proposals/0456-stdlib-span-properties.md) for Swift 6.2

Fixes #1073

Checklist:
- [x] Tests added.

- rename the `bytes` computed property in the extension on `Data` to `byteArray`.
- this is a breaking change, but nonetheless necessary for compatibility with Xcode 26


c128128:
This PR has broken a large part of my dependency tree.

nathanfallet:
Agree, people were blocked with the new bytes property from Apple so their code could not compile. I don't think there is any other way to solve this issue.

PR #1081, opened 2025-10-17
Use FoundationEssentials when available

lemonmojo:
This PR replaces all occurrences of 

```swift
import Foundation
```

with

```swift
#if canImport(FoundationEssentials)
import FoundationEssentials
#else
import Foundation
#endif
```

The benefit is that for platforms that do support FoundationEssentials (ie. non-Apple platforms), it results in less Swift library dependencies and ultimately in a smaller file size of the final distributable package.

There are two additional changes in this PR that could be removed if so desired:
1. Adds `workflow_dispatch` to the CI yml files to allow manually triggered runs. (This was useful for testing in the forked repo)
2. Fixes an issue (probably with newer Swift/Xcode versions) that lead to a compilation issue because the reserved keyword `unsafe` was used in [RSASecKeyTests.swift](https://github.com/krzyzanowskim/CryptoSwift/compare/main...royalapplications:CryptoSwift:foundationessentials?expand=1#diff-946480c24138d055083a8ffef0fb4946df218b7fe388ace83009647c1ed33832).

All tests completed successfully locally on macOS using either Xcode or `swift test` and on Github's CI machines.

Update 1:
So to check if libFoundation.so is actually not(!) linked in on Linux I switched the output type of the target to `.dynamic`, did a `swift build` and `objdump -p libCryptoSwift.so`. Unfortunately it was still there. So I investigated and found out that this is caused by copying the PrivacyInfo.xcprivacy file. It very much sounds like a Swift/SPM bug to me. As soon as I remove the resource, only libFoundationEssentials.so is linked, not libFoundation.so. So I made another change, this time to Package.swift to exclude that file when compiling for non-Apple platforms. Unfortunately, [the change](https://github.com/krzyzanowskim/CryptoSwift/compare/main...royalapplications:CryptoSwift:foundationessentials?expand=1#diff-f913940c58e8744a2af1c68b909bb6383e49007e6c5a12fb03104a9006ae677e) is pretty hacky as it will not work when cross-compiling from macOS to Linux for instance.
If anyone knows why this is happening or has a better idea on how to handle this, I'm all ears.

Update 2:
[Refactored](https://github.com/krzyzanowskim/CryptoSwift/pull/1081/commits/a59b4d91ebb22011656c830f874fe7152e183a57) the resources to be contained in a separate target which is added as a dependency of CryptoSwift only for Apple platforms. This gets rid of the cross-compilation issue from Update 1.


krzyzanowskim:
great idea. Applied upstream with minor fixes.

PR #1083, opened 2025-12-18
Enable LibraryEvolution since the library is stable and avoid ABI bre…

bigbangvn:
Enable LibraryEvolution since the library is stable and avoid ABI breaking issue when an xcframework uses SwiftyRSA

Fixes #

Checklist:
- [ ] Tests added.

-

Enable LibraryEvolution since the library is stable and avoid ABI breaking issue when an xcframework uses SwiftyRSA

Example.

- MyLibrary is compiled to xcframework and it uses SwiftyRSA version x.1
- MyLibrary is used in application A, that uses SwiftyRSA version x.2
- SwiftyRSA version x.2 only reorders/inserts a new public function to a class, but without Library Evolution on, that can cause ABI breaking issue -> crash at runtime of application A, unexpected behaviour.

Benefits for prebuilt xcframework:

- ABI resilient
- Xcode/swift version compatibility

### DIFF
--- a/CryptoSwift.podspec
+++ b/CryptoSwift.podspec
@@ -17,4 +17,5 @@ Pod::Spec.new do |s|
   s.source_files  = "Sources/CryptoSwift/**/*.swift"
   s.requires_arc = true
   s.resource_bundles = {'CryptoSwift' => ['Sources/CryptoSwift/PrivacyInfo.xcprivacy']}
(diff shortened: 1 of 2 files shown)


krzyzanowskim (commented):
per docs https://developer.apple.com/documentation/packagedescription/swiftsetting/unsafeflags%28_%3A_%3A%29

> the use of unsafe flags makes the products containing this target ineligible for use by other packages.

PR #1084, opened 2025-12-25
Change IV length range for XChaCha20 cipher

james-kwong:
Updated the valid range of initialization vector lengths for the XChaCha20 cipher from 12 bytes to 24 bytes, which is the correct value.

Checklist:
- [x] Tests added.


krzyzanowskim:
why? this need more merit to discuss

krzyzanowskim (approved):

